PR #1, opened 2019-07-15
CustomLookupSkill

kargosh123:
## Purpose

* Local Testing of a CustomLookup Skill. The azure function attempts to solve the named entity issue as based on TextAnalytics no user-defined entities are recognized. Currently, there is no fuzziness when comparing entities with the documents and only Latin-based characters are compared.

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[x ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone https://github.com/kargosh123/azure-search-power-skills.git
```

* Test the code

```
Open CustomLookup solution and run the Azure Function in "CustomEntitySearch.cs"
Open CustomLookupTest solution and run tests in "test.cs"
```

## What to Check
* All tests are successful and encompass the potential edge cases for this skill.


CustomLookup/.gitignore
```diff
@@ -0,0 +1,264 @@
+## Ignore Visual Studio temporary files, build results, and
```
bleroy: There's already a gitignore at the root of the repo, you can remove this one.

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
+using System;
```
bleroy: The repo organizes skills in Text/Geo/Vision folders. Looks like this one should be moved under Text.

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using System.Net.Http.Formatting;
+using System.Text;
+using Microsoft.AspNetCore.Routing.Template;
+
+namespace CustomLookup
```
bleroy: You can look at other skills to see how to structure your namespace. It should probably be `AzureCognitiveSearch.PowerSkills.Text.CustomEntitySearch`

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+    public static class CustomEntitySearch
+    {
+        /// <summary>
+        /// We assert the following assumptions:
+        /// 1. All text files are Latin based (i.e. no languageCode)
+        /// 2. Words can contain special characters and numbers
+        /// 3. The provided entities are not case sensitive
+        /// </summary>
+        #region Classes used to deserialize the request
+        private class InputRecord
```
bleroy: Other skills in the repo use a common set of utilities and structures that promote re-use and reduce the clutter in each skill. Look at existing skills and refactor accordingly. First step is to adjust the csproj file to include those utilities.

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+                }
+
+                OutputRecord responseRecord = new OutputRecord
+                {
+                    RecordId = record.RecordId
+                };
+
+                try
+                {
+                    responseRecord.Data = await FindEntity(record.Data.Text, record.Data.Words);
```
bleroy: Once you refactor to use the utility classes, this should become the heart of this method, with little plumbing around it. Look at other skills in the repo for reference.

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+            for (int i = 0; i < words.Count; i++)
+            {
+                string alteredWord = Regex.Escape(words[i]);
+                multiComp[i] = alteredWord;
+            }
+
+            for (int j = 0; j < words.Count; j++)
+            {
+                string pattern = @"\b(?ix-m:" + multiComp[j] + ")";
+                Match entityMatch = Regex.Match(text, pattern, RegexOptions.IgnoreCase, new TimeSpan(1));
```
bleroy: If you're going to use the overload that takes a timeout, make it more than a tick, and move it to a constant.

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+            string[] multiComp = new string[words.Count];
+            for (int i = 0; i < words.Count; i++)
+            {
+                string alteredWord = Regex.Escape(words[i]);
+                multiComp[i] = alteredWord;
+            }
+
+            for (int j = 0; j < words.Count; j++)
+            {
+                string pattern = @"\b(?ix-m:" + multiComp[j] + ")";
```
bleroy: Wouldn't you also want a word boundary after the word?

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// </summary>
+        /// <param name="text"> Text document provided by user</param>
+        /// <param name="words"> User-defined entities </param>
+        /// <returns></returns>
+        private async static Task<List<OutputRecord.OutputRecordData>> FindEntity(string text, List<string> words)
+        {
+            List <OutputRecord.OutputRecordData>  data = new List<OutputRecord.OutputRecordData>();
+
+            string[] multiComp = new string[words.Count];
+            for (int i = 0; i < words.Count; i++)
```
bleroy: You could simplify that with `string[] escapedWords = words.Select(Regex.Escape).ToArray()`. (I also changed the name of that variable to something a little more expressive.

bleroy: Actually, there's no reason to allocate and keep this array around at all. You can escape each word as part of your second loop.

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                string alteredWord = Regex.Escape(words[i]);
+                multiComp[i] = alteredWord;
+            }
+
+            for (int j = 0; j < words.Count; j++)
+            {
+                string pattern = @"\b(?ix-m:" + multiComp[j] + ")";
+                Match entityMatch = Regex.Match(text, pattern, RegexOptions.IgnoreCase, new TimeSpan(1));
+                bool success = (multiComp[j] == "" && text != "") ? false : entityMatch.Success;
```
bleroy: `text == ""` should be handled as the first thing you do in the method (use `string.IsNullOrWhitespace` and return `List<OutputRecordData>.Empty`). Also, test `string.IsNullOrEmpty(words[j])` before you even build the regex, to avoid doing unnecessary work.

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+            List <OutputRecord.OutputRecordData>  data = new List<OutputRecord.OutputRecordData>();
+
+            string[] multiComp = new string[words.Count];
+            for (int i = 0; i < words.Count; i++)
+            {
+                string alteredWord = Regex.Escape(words[i]);
+                multiComp[i] = alteredWord;
+            }
+
+            for (int j = 0; j < words.Count; j++)
```
bleroy: Favor `foreach` here since you don't need the index.

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                string pattern = @"\b(?ix-m:" + multiComp[j] + ")";
+                Match entityMatch = Regex.Match(text, pattern, RegexOptions.IgnoreCase, new TimeSpan(1));
+                bool success = (multiComp[j] == "" && text != "") ? false : entityMatch.Success;
+                if (success)
```
bleroy: This can be made terser:

```csharp
data.Add(
    new OutputRecord.OutputRecordData{
        Name = words[j],
        MatchIndex = success ? entityMatch.Index : -1
    }
);
```

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+        #endregion
+
+        #region Classes used to serialize the response
+
+        private class OutputRecord
+        {
+            public class OutputRecordData
+            {
+                public string Name { get; set; } = "";
+                public string MatchIndex { get; set; } = "";
```
bleroy: Why is that a string and not an int?

CustomLookup/CustomLookup.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <AzureFunctionsVersion>v2</AzureFunctionsVersion>
+  </PropertyGroup>
```
bleroy: Also set the default namespace here.

CustomLookup/CustomLookup.csproj
```diff
@@ -0,0 +1,19 @@
...
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="1.0.29" />
+    <PackageReference Include="Moq" Version="4.12.0" />
```
bleroy: Why is the main project, not the test project, referencing Moq?

CustomLookup/samplefile.txt
```diff
@@ -0,0 +1,10 @@
+Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam ut dictum ligula. Aliquam eget viverra sapien. Sed vitae mi sit amet ex laoreet pellentesque eget nec risus. Morbi metus sem, pharetra eu pretium in, mollis et ante. Etiam non leo quis massa finibus vehicula et vitae est. Ut id elementum nisi, nec mollis diam. Suspendisse iaculis, velit eu porta placerat, sem arcu varius est, sit amet tincidunt lectus dolor in nunc.
```
bleroy: There's a top-level directory for sample files.

CustomLookup/CustomLookup.sln
```diff
@@ -0,0 +1,30 @@
+﻿
```
bleroy: Please integrate with the top-level solution, instead of creating your own.

Vision/ImageStore/README.md
```diff
@@ -5,7 +5,6 @@ languages:
 products:
 - azure-cognitive-services
 name: Image store sample skills for cognitive search
-urlFragment: azure-imagestore-sample
```
bleroy: Please revert that.

CustomLookupTest/test.cs
```diff
@@ -0,0 +1,286 @@
...
+using System.Linq;
+using System.Runtime.ExceptionServices;
+/// <summary>
+/// This is my attempt at a unit test. I took information from VSE documentation and the team's current SkillsetsTest.cs
+/// </summary>
+namespace Test
+{
+
+    [TestClass]
+    public class CustomLookupTests
```
bleroy: It seems like your tests are a little more than unit tests: there should be no need to call your function using a web request. By doing so, you're testing the Azure functions infrastructure rather than just your code. I'd re3factor to just pass data into your function directly.

CustomLookup/CustomLookup.csproj
bleroy: Actually, it seems like your tests aren't using Moq at all, so you should be able to remove this entirely.

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
+using System;
+using System.Collections.Generic;
```
jadrefke: @ do these files need MSFT copyright headers?

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                public string Text { get; set; }
+                public List<string> Words { get; set; }
+            }
+
+            public string RecordId { get; set; }
+            public InputRecordData Data { get; set; }
+        }
+
+        private class WebApiRequest
```
jadrefke: >class [](start = 16, length = 5)

If you end up keeping these models (rather than using common util models) please move each class to its own file

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+using System.Net.Http.Formatting;
+using System.Text;
+using Microsoft.AspNetCore.Routing.Template;
+
+namespace CustomLookup
+{
+    /// <summary>
+    /// Based on sample custom skill provided in Azure Search. Provided a user-defined list of entities
+    /// this function determines the entities first occurrence within a given document. This list of entities
+    /// must repeatedly be provided by the user for each document.
```
jadrefke: There's no way to supply per-document skill parameters in Cognitive Search pipeline that i can think of. Can you refactor this so that list of entities is a payload level entity rather than document level entity?

CustomLookup/CustomEntitySearch.cs
```diff
@@ -0,0 +1,187 @@
...
+                {
+                    response.Values.Add(responseRecord);
+                }
+            }
+
+            return (ActionResult)new OkObjectResult(response);
+        }
+        #endregion
+
+        #region Helper functions (My main change to the sample code provided)
```
jadrefke: >(My main change to the sample code provided) [](start = 33, length = 44)

nit: can remove this comment

CustomLookup/CustomEntitySearch.cs
jadrefke: eg: TimeSpan.FromSeconds(1) 

---
In reply to: [303662654](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r303662654) [](ancestors = 303662654)

CustomLookupTest/Constants.cs
```diff
@@ -0,0 +1,190 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Tests
```
jadrefke: >Tests [](start = 10, length = 5)

update default and file namespaces for this project as well

CustomLookup/CustomEntitySearch.cs
jadrefke: >x-m [](start = 40, length = 3)

do you need the x or -m group modifiers?

CustomLookupTest/Definitions.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Tests
+{
+    public class SkillInput
```
jadrefke: >SkillInput [](start = 17, length = 10)

reuse models from main project (or util project per bertrand's comments)

CustomLookup/samplefile.txt
jadrefke: is this used anywhere?

---
In reply to: [303668365](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r303668365) [](ancestors = 303668365)

CustomLookupTest/test.cs
```diff
@@ -0,0 +1,286 @@
...
+using System.Linq.Expressions;
+using System.Reflection.Metadata;
+using System.Threading;
+using System.Timers;
+using System.Linq;
+using System.Runtime.ExceptionServices;
+/// <summary>
+/// This is my attempt at a unit test. I took information from VSE documentation and the team's current SkillsetsTest.cs
+/// </summary>
+namespace Test
```
jadrefke: >Test [](start = 10, length = 4)

name this file something more descriptive? eg: LookupTests for pattern matching unit tests or WebResponseTests for integration tests validating response codes

CustomLookupTest/Constants.cs
```diff
@@ -0,0 +1,190 @@
...
+{
+    public class Constants
```
jadrefke: >Constants [](start = 17, length = 9)

nit: can you rename this class to something like TestData?

CustomLookupTest/test.cs
```diff
@@ -0,0 +1,286 @@
...
+            string responseString = response.Content.ReadAsStringAsync().Result;
+            Assert.AreEqual("Bad Request", response.ReasonPhrase, false);
+            Assert.AreEqual(Constants.checkTest1, responseString, false);
+        }
+
+        [TestMethod]
+        public void CustomLookupTest2()
+        {
+            // tests against incorrect input (missing text)
+            string content = Constants.inputTest2;
```
jadrefke: >inputTest2 [](start = 39, length = 10)

use descriptive names even for test variables. eg: MissingTextPayload

CustomLookupTest/test.cs
```diff
@@ -0,0 +1,286 @@
...
+            string content = Constants.inputTest1;
+            HttpContent jsonContent = new StringContent(content, null, "application/json");
+            var response = client.PostAsync("http://localhost:7071/api/CustomEntitySearch", jsonContent).Result;
+            string responseString = response.Content.ReadAsStringAsync().Result;
+            Assert.AreEqual("Bad Request", response.ReasonPhrase, false);
+            Assert.AreEqual(Constants.checkTest1, responseString, false);
+        }
+
+        [TestMethod]
+        public void CustomLookupTest2()
```
jadrefke: >CustomLookupTest2 [](start = 20, length = 17)

Use descriptive name for test methods, eg: MissingTextShouldReturnBadRequest

CustomLookupTest/test.cs
```diff
@@ -0,0 +1,286 @@
...
+            Assert.AreEqual(Constants.checkTest2, responseString, false);
+        }
+
+        [TestMethod]
+        public void CustomLookupTest3()
+        {
+            // tests against empty string text
+            string text = Constants.inputElement.Replace("#REPLACE ME#", @"""""");
+            string words = text.Replace("#INSERT WORDS#", Constants.inputTest3);
+            string content = Constants.inputCheckTest.Replace("#REPLACE ME#", words);
```
jadrefke: >REPLACE ME [](start = 64, length = 10)

nitish: it might be more maintainable to refactor these into a helper methods

eg:

public static string GetContentPaylaod(string words)
{
    return $@"{{""values"":[{words}]}}"
}

CustomLookup/CustomEntitySearch.cs
jadrefke: specifically the x modifier seems particularly restrictive. Like if the user wanted to search for "George Washington"

---
In reply to: [304008744](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r304008744) [](ancestors = 304008744)

CustomLookup/CustomEntitySearch.cs
bleroy: I think for open source projects, it's not required (see release checklist on the internal open source site). Doesn't hurt to add it though, I'll check for more authoritative advice and do it across the repo if it's required, no need to address it as part of this PR.

bleroy: Doesn't look like we *need* copyright headers, but it's generally considered good practice. See for instance https://forums.dotnetfoundation.org/t/file-headers-and-copyright-statements/1276 for the .NET Foundation's stance on the issue. I'll add notices throughout the repo.

CustomLookup/.gitignore
kargosh123: Removed

CustomLookup/CustomEntitySearch.cs
kargosh123: Moved to Text

CustomLookup/samplefile.txt
kargosh123: No sorry,  it was used for testing initially.

CustomLookup/CustomEntitySearch.cs
kargosh123: Currently, I can't find a way to have the list of entities at a payload level as each text has a recordId

CustomLookup/CustomEntitySearch.cs
kargosh123: the x modifier only ignores unescaped whitespace characters, so it should be ok.

CustomLookupTest/test.cs
kargosh123: I am still unsure how I am supposed to pass data into my function directly without using Moq or some "substitute" client. If possible, could you explain how I should go about running my tests this way?

CustomLookup/CustomEntitySearch.cs
kargosh123: sorry, i saw this as a string in an example but realized i made a mistake. changed to int

CustomLookup/CustomEntitySearch.cs
kargosh123: currently, i thought that wrt english, there are more cases where we want the entity to imply the same meaning as found in the text, which would mean that different suffixes would be okay (ex. conscientious vs. conscientiousness). If we only want exact cases, then I should put the word boundary.

CustomEntitySearchTest/CustomEntitySearchTest.sln
```diff
@@ -0,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
```
jadrefke: Don't think this one needs its own solution. Recommend you either create a Tests folder in the existing PowerSkills sln, and put this project there, or you can create a new top level PowerSkillTests.sln

CustomEntitySearchTest/Constants.cs
```diff
@@ -0,0 +1,212 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace LookupTests
+{
+    public class TestData
```
jadrefke: >TestData [](start = 17, length = 8)

rename file to match the class name

CustomEntitySearchTest/Constants.cs
```diff
@@ -0,0 +1,212 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace LookupTests
```
jadrefke: >LookupTests [](start = 10, length = 11)

AzureCognitiveSearch.PowerSkills.Text.CustomEntitySearch.Tests

Text/CustomEntitySearch/azuredeploy.json
```diff
@@ -0,0 +1,118 @@
...
+        ],
+        "metadata": {
+          "description": "Storage Account type"
+        }
+      }
+  },
+  "variables": {
+    "repoURL": "https://github.com/Azure-Samples/azure-search-power-skills",
+    "functionProject": "Text\\BingEntitySearch\\BingEntitySearch.csproj",
+    "websiteName": "[toLower(concat(parameters('resourcePrefix'), '-site-', uniqueString(resourceGroup().id)))]",
```
jadrefke: project needs to be for this one

Text/CustomEntitySearch/Entities.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace AzureCognitiveSearch.PowerSkills.Text.CustomEntitySearch
+{
+    public class Entities
```
jadrefke: >Entities [](start = 17, length = 8)

This looks like a singular Entity, not plural Entities

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -0,0 +1,75 @@
...
+            log.LogInformation("Custom Entity Search function: C# HTTP trigger function processed a request.");
+
+            string skillName = executionContext.FunctionName;
+            IEnumerable<WebApiRequestRecord> requestRecords = WebApiSkillHelpers.GetRequestRecords(req);
+            if (requestRecords == null)
+            {
+                return new BadRequestObjectResult($"{skillName} - Invalid request record array.");
+            }
+
+            const int MAXTIME = 1;
```
jadrefke: >MAXTIME [](start = 22, length = 7)

nit: use descriptive variable names. EG: MaxRegexEvaluationTime

jadrefke: nit: can also move this to a static class const rather than defining in function body

---
In reply to: [304663420](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r304663420) [](ancestors = 304663420)

Text/CustomEntitySearch/CustomLookup.sln
```diff
@@ -0,0 +1,30 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
```
jadrefke: can you remove this sln file?

Text/CustomEntitySearch/README.md
```diff
@@ -0,0 +1,116 @@
...
+    "batchSize": 1,
+    "inputs": [
+        {
+            "name": "name",
+            "source": "/document/merged_content/*"
+        }
+    ],
+    "outputs": [
+        {
+            "name": "description",
```
jadrefke: >description [](start = 21, length = 11)

don't think you put the data in the description field

Text/CustomEntitySearch/samplefile.txt
```diff
@@ -0,0 +1,10 @@
+Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam ut dictum ligula. Aliquam eget viverra sapien. Sed vitae mi sit amet ex laoreet pellentesque eget nec risus. Morbi metus sem, pharetra eu pretium in, mollis et ante. Etiam non leo quis massa finibus vehicula et vitae est. Ut id elementum nisi, nec mollis diam. Suspendisse iaculis, velit eu porta placerat, sem arcu varius est, sit amet tincidunt lectus dolor in nunc.
```
jadrefke: >Lorem [](start = 0, length = 5)

this unused?

Text/CustomEntitySearch/WebAPISkillContract.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections.Generic;
+
+namespace AzureCognitiveSearch.PowerSkills.Common
+{
```
jadrefke: don't copy these files here, just link to them

jadrefke: same for WebAPISkillHelper.cs

---
In reply to: [304669534](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r304669534) [](ancestors = 304669534)

CustomEntitySearchTest/Constants.cs
jadrefke: update default namespace, and namespaces of other files as well

---
In reply to: [304660280](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r304660280) [](ancestors = 304660280)

CustomEntitySearchTest/Entities.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace AzureCognitiveSearch.PowerSkills.Text.CustomEntitySearch
+{
+    public class Entities
+    {
+        public string Name { get; set; }
+        public int MatchIndex { get; set; }
+    }
```
jadrefke: reference the other project to utilize this model

CustomEntitySearchTest/Definitions.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Tests
+{
+    public class SkillInput
+    {
```
jadrefke: these classes look unused, can you delete this file?

CustomEntitySearchTest/WebAPISkillContract.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections.Generic;
+
```
jadrefke: don't copy these files here, just add a reference to them.
Same for WebAPISkillHelper.cs

CustomEntitySearchTest/LookupTests.cs
```diff
@@ -0,0 +1,191 @@
+using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using System.Net.Http;
+using System.Threading.Tasks;
+using System.Linq;
+using AzureCognitiveSearch.PowerSkills.Common;
+/// <summary>
+/// This is my attempt at a unit test. I took information from VSE documentation and the team's current SkillsetsTest.cs
+/// </summary>
```
jadrefke: unnecessary comment, can you just delete this?

The mentioned file is from another repository as well, so not useful to anyone reading it here

CustomEntitySearchTest/Constants.cs
```diff
@@ -0,0 +1,212 @@
...
+        public static string GetPayload(string inputText, string inputWords)
+        {
+            string text = inputElement.Replace("#REPLACE ME#", inputText);
+            string words = text.Replace("#INSERT WORDS#", inputWords);
```
jadrefke: >words [](start = 19, length = 5)

nit: confusing variable names

CustomEntitySearchTest/LookupTests.cs
```diff
@@ -0,0 +1,191 @@
...
+            string checkTest6 = TestData.GetOutput(TestData.inputTest6words, TestData.outputTest6);
+            Assert.AreEqual(checkTest6, responseString, false);
+        }
+
+        [TestMethod]
+        public void LargeDatasetQuickResult()
+        {
+            //tests against a large number of documents inputted (loadtest)
+            string text = TestData.inputElement.Replace("#REPLACE ME#", TestData.largestText);
+            string words = text.Replace("#INSERT WORDS#", TestData.inputTest5);
```
jadrefke: can you refactor this in with your other helper methods in TestData

CustomEntitySearchTest/Constants.cs
```diff
@@ -0,0 +1,212 @@
...
...1, -1, -1, -1, -1, -1, -1, 6917, -1, -1, -1, -1, 12232, -1, -1, -1, -1, -1, -1, -1, 8522, -1, -1, -1, 2363, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 7856, -1, 3738, -1, -1, -1, 8833, 8892, -1, -1, -1, -1, 5036, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 8290, -1, 8990, 1505, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 2133, -1, 5344, -1, -1, -1, -1, -1, -1, 515, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 422, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 8603, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 8858, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 35, -1, 7419, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 5726, -1, -1, -1, 5691, -1, -1, -1, 3124, -1, -1, -1, -1, -1, -1, -1, -1, 4056, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 3099, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 11826, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 3174, -1, -1, -1, -1, -1, -1, -1, -1, 249";
```
jadrefke: use descriptive variable names

even its just like {TestName}ExpectedInput and {TestName}ExpectedOutput

Text/CustomEntitySearch/README.md
kargosh123: Is it alright if I create a new field (ex. entities)?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -0,0 +1,75 @@
...
+                            data.Add(
+                                new Entity
+                                {
+                                    Name = word,
+                                    MatchIndex = entityMatch.Success ? entityMatch.Index : -1
+                                }) ;
+                        }
+                    }
+
+                    outRecord.Data["Entity"] = data;
```
jadrefke: >Entity [](start = 36, length = 6)

this one i think you would name plural since its a list. Entities

Text/CustomEntitySearch/README.md
```diff
@@ -0,0 +1,116 @@
...
+    "batchSize": 1,
+    "inputs": [
+        {
+            "name": "name",
+            "source": "/document/merged_content/*"
+        }
+    ],
+    "outputs": [
+        {
+            "name": "entities",
```
jadrefke: >entities [](start = 21, length = 8)

you changed the name of this in CustomEntitySearch it seems

CustomLookupTest/test.cs
jadrefke: You should be able to just call directly into the function. 
EG:
CustomEntitySearch.Run(TestData.GeneratePayloadRequest(), new TestLogger(), new ExecutionContext(~~,~~,~))

---
In reply to: [304646902](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r304646902) [](ancestors = 304646902)

Tests/CustomEntitySearchTest/CustomEntitySearchTest.cs
```diff
@@ -0,0 +1,167 @@
...
+            WebApiResponseRecord output = new WebApiResponseRecord();
+            try
+            {
+                output = JsonConvert.DeserializeObject<WebApiResponseRecord>(responseString);
+            }
+            catch
+            {
+                Assert.Fail("Skill failed to handle an empty test. Errored out.");
+            }
+            string checkTest5 = TestData.GetOutput(TestData.largeTextQuickResultInputWords, TestData.largeTextQuickResultExpectedOutput);
```
jadrefke: >checkTest5 [](start = 19, length = 10)

try to use descriptive variable names even for tmp objects

Tests/CustomEntitySearchTest/CustomEntitySearchTest.cs
```diff
@@ -0,0 +1,167 @@
...
+                output = JsonConvert.DeserializeObject<WebApiResponseRecord>(responseString);
+            }
+            catch
+            {
+                Assert.Fail("Skill failed to handle an empty test. Errored out.");
+            }
+            string checkTest8 = TestData.GetOutput(TestData.largestWords, TestData.largeNumWordsQuickResultExpectedOutput);
+            Assert.AreEqual(checkTest8, responseString, false);
+        } 
+    }
```
jadrefke: Would be useful to write more tests around word delineation. For example, it looks like the pattern "car" would currently match a document containing "We'll perform a cardiovascular operation at noon."

Text/CustomEntitySearch/CustomEntitySearch.cs
bleroy: Yes, and the `data` variable could be more expressively named `entities` as well.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                return new BadRequestObjectResult($"{skillName} - Invalid request record array.");
+            }
+
+            WebApiSkillResponse response = WebApiSkillHelpers.ProcessRequestRecords(skillName, requestRecords,
+                (inRecord, outRecord) => {
+                    string text = inRecord.Data["text"] as string;
+                    List<string> words = ((JArray)inRecord.Data["words"]).ToObject<List<string>>();
+
+                    List<Entity> data = new List<Entity>();
```
bleroy: When newing up an object, `var` is preferred because the type is unambiguous and that's more DRY.

Text/CustomEntitySearch/README.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+## Sample Output:
+
+```json
+{
+    "values": [
+        {
+            "recordId": "1",
+            "data": {
```
bleroy: This brace is never closed, making the document invalid.

Text/CustomEntitySearch/README.md
```diff
@@ -0,0 +1,116 @@
...
+						"matchIndex": 45
+					}
+			]
+        },
+        {
+            "recordId": "foo1",
+            "data": 
+			{
+				"entities": [
+					{
```
bleroy: Please make sure indentation style is consistent internally (you have a mix of tabs and spaces) and with other samples' READMEs.

Text/CustomEntitySearch/README.md
```diff
@@ -0,0 +1,116 @@
...
+```json
+{
+    "@odata.type": "#Microsoft.Skills.Custom.WebApiSkill",
+    "description": "Our Custom Entity search custom skill",
+    "context": "/document/merged_content/*",
+    "uri": "[AzureFunctionEndpointUrl]/api/custom-search?code=[AzureFunctionDefaultHostKey]",
+    "batchSize": 1,
+    "inputs": [
+        {
+            "name": "name",
```
bleroy: You should have "words" and "text" input parameters instead of the "name".

Text/CustomEntitySearch/azuredeploy.json
```diff
@@ -0,0 +1,118 @@
...
+      ],
+      "properties": {
+        "name": "[variables('functionAppName')]",
+        "kind": "functionapp",
+        "httpsOnly": true,
+        "siteConfig": {
+          "appSettings": [
+            {
+              "name": "WEBSITE_CONTENTAZUREFILECONNECTIONSTRING",
+              "value": "[concat('DefaultEndpointsProtocol=https;AccountName=', variables('storageAccount'), ';AccountKey=', listKeys(resourceId('CogSearch','Microsoft.Storage/storageAccounts', variables('storageAccount')), '2015-06-15').key1)]"
```
bleroy: For this setting and the next two, remove `"CogSearch",`. This is a mistake I made in my own deployment files that I fixed on Tuesday in all existing skills.

CustomLookupTest/test.cs
bleroy: Add a project reference to the main project from your test project, and call `CustomEntitySearch.Run` directly instead of doing a `PostRequest`.

bleroy: You may also want to add a helper method to your test project that builds the `HttpRequest` from text and words, which should remove a lot of the repetition in those tests.

CustomLookup/CustomEntitySearch.cs
bleroy: Couldn't you use the output of entity detection, for example, be the input for your words parameter?

Tests/CustomEntitySearchTest/CustomEntitySearchTest.cs
kargosh123: Yes, it would. I changed it to be exact pattern matching for now.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -50,10 +50,15 @@ public static class CustomEntitySearch
 
             WebApiSkillResponse response = WebApiSkillHelpers.ProcessRequestRecords(skillName, requestRecords,
                 (inRecord, outRecord) => {
-                    string text = inRecord.Data["text"] as string;
+                    string text = Regex.Escape(inRecord.Data["text"] as string);
```
bleroy: Why do you need to escape this?

Text/CustomEntitySearch/words.json
```diff
@@ -0,0 +1,3 @@
+{
+  "targetWords": ["foo1", "foo2"]
```
bleroy: Since the file only contains a list of words, the outer object is not necessary, you can have the array at the top level directly. Or we could make it even easier for the customer and have a text file with one word per line. Less verbose.

Text/CustomEntitySearch/README.md
```diff
@@ -21,7 +21,14 @@ These skills have no additional requirements than the ones described in [the roo
 
 ## Settings
 
-This function requires Latin-based text (as seen in the sample document provided).
+This function requires Latin-based text (as seen in the sample document provided). The input field "words" is optional, where a user can add a "words.json" file instead.
+
+## Sample Config File
+```json
+{
+	["foo1", "foo2"]
```
bleroy: You forgot the "targetWords" property name. Then again, see my comment below on simplifying the format of the file.

Text/CustomEntitySearch/README.md
```diff
@@ -109,10 +115,18 @@ Here's a sample skill definition for this example (inputs and outputs should be
         {
             "name": "words",
             "source": "/document/merged_content/*"
+        },
+		{
+            "name": "languageCode",
+            "source": "/document/merged_content/*"
```
bleroy: This path doesn't look right. @jadrefke can you take another look at this skill definition and advise?

Text/CustomEntitySearch/WordLinker.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.IO;
+using System.Text;
+
+namespace AzureCognitiveSearch.PowerSkills.Text.CustomEntitySearch
+{
+    class WordsConfig
+    {
+        public List<string> TargetWords { get; set; }
+    }
+    class WordLinker
```
bleroy: The `WordsConfig` class doesn't seem to be very useful. You should be able to deserialize your file directly to a `List<string>`.

bleroy: Also make that class explicitly public or internal. (I know I forgot that on the `AcronymLinker` class, which I'll fix now).

Text/CustomEntitySearch/README.md
jadrefke: default output location for [language detection skill](https://docs.microsoft.com/en-us/azure/search/cognitive-search-skill-language-detection) language code is /document/languageCode

---
In reply to: [305524700](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r305524700) [](ancestors = 305524700)

jadrefke: text's source should probably be /document/content

words source would likely be something custom.. Another option would be to include the list of words in a custom webApiHeader... then you don't need to grab it from each individual document

---
In reply to: [305533021](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r305533021) [](ancestors = 305533021,305524700)

Tests/CustomEntitySearchTest/CustomEntitySearchTest.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
+    <PackageReference Include="MSTest.TestAdapter" Version="1.3.2" />
+    <PackageReference Include="MSTest.TestFramework" Version="1.3.2" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="../../*.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="../../Text\CustomEntitySearch\Entity.cs" />
+  </ItemGroup>
```
jadrefke: Usually you add a reference to the entire CSProj, eg:
    <ProjectReference Include="..\..\Text\CustomEntitySearch\CustomEntitySearch.csproj" />

Text/CustomEntitySearch/WordLinker.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public WordLinker(string executingDirectoryPath)
+        {
+            string json = File.ReadAllText($"{executingDirectoryPath}\\words.json");
+            Words = new List<string>(
+                JsonConvert.DeserializeObject<WordsConfig>(json).TargetWords);
+        }
+
+        public List<string> Words
```
jadrefke: >List [](start = 15, length = 4)

nit: IList

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -0,0 +1,93 @@
...
+                (inRecord, outRecord) => {
+                    string text = Regex.Escape(inRecord.Data["text"] as string);
+                    List<string> words = ((JArray)inRecord.Data["words"]).ToObject<List<string>>();
+                    if (words == null)
+                    {
+                        words = new WordLinker(executionContext.FunctionAppDirectory).Words;
+                    }
+
+                    var entities = new List<Entity>();
+                    var entitiesFound = new List<string>();
```
jadrefke: I might have missed some context, but why are you returning two lists now?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// this function determines the Entity first occurrence within a given document. This list of entities
+    /// must repeatedly be provided by the user for each document.
+    /// </summary>
+    public static class CustomEntitySearch
+    {
+        private static int MaxRegexEvalTime = 1;
+        /// <summary>
+        /// We assert the following assumptions:
+        /// 1. All text files are Latin based (i.e. no languageCode)
+        /// 2. Words can contain special characters and numbers
```
jadrefke: this still true?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -0,0 +1,93 @@
...
+            IEnumerable<WebApiRequestRecord> requestRecords = WebApiSkillHelpers.GetRequestRecords(req);
+            if (requestRecords == null)
+            {
+                return new BadRequestObjectResult($"{skillName} - Invalid request record array.");
+            }
+
+            WebApiSkillResponse response = WebApiSkillHelpers.ProcessRequestRecords(skillName, requestRecords,
+                (inRecord, outRecord) => {
+                    string text = Regex.Escape(inRecord.Data["text"] as string);
+                    List<string> words = ((JArray)inRecord.Data["words"]).ToObject<List<string>>();
```
jadrefke: >Data["words"] [](start = 59, length = 13)

where are you using languageCode from the readme?

kargosh123: Currently not using it. It most likely will have to be used for fuzzy matching.

Tests/CustomEntitySearchTest/CustomEntitySearchTest.csproj
bleroy: Yes, to be clear, the reason why I made a file include for the utility classes was so each project can be as self-contained as possible while still using common logic. In this case, it's entirely different though, and it should be a `ProjectReference`.

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: According to the emails, they want a list of all entities found as well as a more "detailed" version with the matchIndex and later the fuzzy correlation score

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: I am no longer working under the assumption that the characters will only be latin-based

Text/CustomEntitySearch/CustomEntitySearch.cs
bleroy: So the issue here will be that you'll have duplicate items in `entitiesFound`. `entities` should be a list of matches (with index), wehreas `entitiesFound` should be a list of the found words with no duplicates. You could use a `HashSet` for the latter to easily get rid of the duplicate entries, and just `ToArray` it when you're done.

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: No, sorry. The files do not only need to be latin based for exact match based on a new assumption that the files are 64 encoded

Text/CustomEntitySearch/words.json
kargosh123: Currently, the words can include new line characters. Do we not want to support multi-line entities?

Tests/CustomEntitySearchTest/CustomEntitySearchTest.csproj
kargosh123: Would that cause any dependency issues when running CustomEntitySearch?

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: Never mind, sorry for this PR this is unnecessary.

Text/CustomEntitySearch/README.md
kargosh123: For now, I'll keep words as is wrt the skill definition so as to maintain a general webAPIHeader. For prod dev, I think it would definitely be better to have a custom one though

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -66,15 +67,17 @@ public static class CustomEntitySearch
                             if (string.IsNullOrEmpty(word)) continue;
                             string escapedWord = Regex.Escape(word);
                             string pattern = @"\b(?ix:" + escapedWord + @")\b";
-                            Match entityMatch = Regex.Match(text, pattern, RegexOptions.IgnoreCase, TimeSpan.FromSeconds(MaxRegexEvalTime));
-                            if (entityMatch.Success)
+                            MatchCollection entityMatch = Regex.Matches(text, pattern, RegexOptions.IgnoreCase, TimeSpan.FromSeconds(MaxRegexEvalTime));
+                            if (entityMatch.Count != 0)
                             {
-                                entities.Add(
+                                entityMatch.OfType<Match>()
```
bleroy: Simpler:
```csharp
foreach(Match match in entityMatch) {
    entities.Add(
        new Entity
        {
            Name = word,
            MatchIndex = match.Index
        });
}
```

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -50,11 +51,11 @@ public static class CustomEntitySearch
 
             WebApiSkillResponse response = WebApiSkillHelpers.ProcessRequestRecords(skillName, requestRecords,
                 (inRecord, outRecord) => {
-                    string text = Regex.Escape(inRecord.Data["text"] as string);
+                    string text = inRecord.Data["text"] as string;
                     List<string> words = ((JArray)inRecord.Data["words"]).ToObject<List<string>>();
                     if (words == null)
                     {
-                        words = new WordLinker(executionContext.FunctionAppDirectory).Words;
+                        words = new List<string>(new WordLinker(executionContext.FunctionAppDirectory).Words);
```
bleroy: No need to allocate a new list: this is already what you're getting out of `Words`.

CustomLookup/CustomEntitySearch.cs
kargosh123: In that case, all texts are under one recordId. Is that okay?

Tests/CustomEntitySearchTest/TestData.cs
```diff
@@ -0,0 +1,274 @@
...
+using System.IO;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace AzureCognitiveSearch.PowerSkills.Tests.CustomEntitySearchTest
+{
+    public class TestData
+    {
+        public const string hostAddress = "http://localhost:7071/api/CustomEntitySearch";
```
jadrefke: >hostAddress [](start = 28, length = 11)

this used anymore?

Tests/CustomEntitySearchTest/TestData.cs
```diff
@@ -0,0 +1,274 @@
...
+        public const string missingWordsBadRequestInput = @"{
+    ""values"": [
+        {
+            ""recordId"": ""1"",
+            ""data"":
+            {
+                ""text"":  ""Pablo Picasso"",
+                ""languageCode"": ""en""
+            }
```
jadrefke: nit: remove languageCode for now since its not used

jadrefke: same for missingTextBadRequestInput below

---
In reply to: [307376004](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r307376004) [](ancestors = 307376004)

Text/CustomEntitySearch/CustomEntitySearch.cs
jadrefke: this is still a list in iteration 10

---
In reply to: [305538620](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r305538620) [](ancestors = 305538620)

Text/CustomEntitySearch/CustomEntitySearch.csproj
```diff
@@ -0,0 +1,30 @@
...
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
+    </None>
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="../../*.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Remove="words.json" />
+  </ItemGroup>
```
jadrefke: you can remove this item group i think

Text/CustomEntitySearch/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Settings
+
+This function requires Latin-based text (as seen in the sample document provided). The input field "words" is optional, where a user can add a "words.json" file instead.
+
+## Sample Config File
+```json
+{
+    ["foo1", "foo2"]
+}
```
jadrefke: words.json appears to be empty in this iteration

Text/CustomEntitySearch/README.md
```diff
@@ -0,0 +1,129 @@
...
+            "data": {
+                "entitiesFound": ["learn", "app"],
+                "entities": [
+                    {
+                        "name": "learn",
+                        "matchIndex": 1
+                    },
+                    {
+                        "name": "app",
+                        "matchIndex": 45
```
jadrefke: can you update this output? i don't think this one should match anymore

Text/CustomEntitySearch/README.md
```diff
@@ -0,0 +1,129 @@
...
+{
+    "@odata.type": "#Microsoft.Skills.Custom.WebApiSkill",
+    "description": "Our Custom Entity search custom skill",
+    "context": "/document/merged_content/*",
+    "uri": "[AzureFunctionEndpointUrl]/api/custom-search?code=[AzureFunctionDefaultHostKey]",
+    "batchSize": 1,
+    "inputs": [
+        {
+            "name": "text",
+            "source": "/document/content/*"
```
jadrefke: >/* [](start = 40, length = 2)

no "/*" at thte end of this path

Text/CustomEntitySearch/README.md
```diff
@@ -0,0 +1,129 @@
...
+        },
+        {
+            "name": "words",
+            "source": "/document/merged_content/*"
+        }
+    ],
+    "outputs": [
+        {
+            "name": "entitiesFound",
```
jadrefke: >entitiesFound [](start = 21, length = 13)

I believe these are case sensitive

PowerSkills.sln
```diff
@@ -44,6 +44,12 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "_SampleData", "_SampleData"
 		SampleData\Invoice_5.pdf = SampleData\Invoice_5.pdf
 	EndProjectSection
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CustomEntitySearch", "Text\CustomEntitySearch\CustomEntitySearch.csproj", "{F0B21155-829F-4B26-853B-82ECAAB18D23}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{F5F3F598-2DCA-42EA-9797-D3A3E7CFA2CB}"
+EndProject
```
jadrefke: remove this project

kargosh123: I thought I fixed this. Tried to just clean again.

jadrefke: its still here in iteration 11

---
In reply to: [307531580](https://github.com/Azure-Samples/azure-search-power-skills/pull/1#discussion_r307531580) [](ancestors = 307531580)

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -0,0 +1,97 @@
...
+                (inRecord, outRecord) => {
+                    string text = inRecord.Data["text"] as string;
+                    IList<string> words = ((JArray)inRecord.Data["words"]).ToObject<List<string>>();
+                    if (words == null)
+                    {
+                        words = new WordLinker(executionContext.FunctionAppDirectory).Words;
+                    }
+
+                    var entities = new List<Entity>();
+                    var entitiesFound = new List<string>();
```
jadrefke: >List [](start = 44, length = 4)

should this be a hash set?

CustomLookup/CustomEntitySearch.cs
bleroy: If you mean the terms to look for, then yes, that is fine as one of the options, and since you added the configurable lookup file as an additional input, I think that would be fine.

Text/CustomEntitySearch/words.json
bleroy: Let's assume not, at least for now.

Text/CustomEntitySearch/CustomEntitySearch.cs
bleroy: Yes.

Text/CustomEntitySearch/README.md
```diff
@@ -26,7 +26,7 @@ This function requires Latin-based text (as seen in the sample document provided
 ## Sample Config File
 ```json
 {
-	["foo1", "foo2"]
+    ["foo1", "foo2"]
```
bleroy: That's still invalid JSON: it's missing a property name for that array.

PR #5, opened 2019-08-05
Fuzzy Matching Abilities (Pre-Prod)

kargosh123:
## Purpose

* Added fuzzy matching and user config abilities

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[x ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone https://github.com/kargosh123/azure-search-power-skills.git
```

* Test the code

```
Open CustomEntitySearchTest project and run tests in "CustomEntitySearchTests.cs"
```

## What to Check
* All tests are successful and encompass the potential edge cases for this skill.


Tests/CustomEntitySearchTest/CustomEntitySearchTest.cs
```diff
@@ -87,5 +89,23 @@ public async Task LargeNumWordsQuickResult()
...
+        // greek, thai, hebrew, turkish, czech, hungarian, arabic, japanese, finnish, danish, norwegian, korean, polish, russian, swedish, japanese (again??), 
+        // italian, portuguese, french, spanish, dutch, german, english
+        [TestMethod]
+        public async Task SupportAllOtherCurrentLanguages()
+        {
+            TestData.supportedTextandWordsTempInitializer();
+            Dictionary<string, string[]> supportedLangTextandWords = TestData.supportedTextandWords;
+            foreach (string key in supportedLangTextandWords.Keys)
+            {
+                Console.WriteLine("Testing language: {0}", key);
```
bleroy: Can you remove those?

Tests/CustomEntitySearchTest/CustomEntitySearchTest.cs
```diff
@@ -87,5 +89,23 @@ public async Task LargeNumWordsQuickResult()
...
             Assert.AreEqual(checkLargeNumWordsQuickResult, outputContent, false);
         }
+        // greek, thai, hebrew, turkish, czech, hungarian, arabic, japanese, finnish, danish, norwegian, korean, polish, russian, swedish, japanese (again??), 
+        // italian, portuguese, french, spanish, dutch, german, english
+        [TestMethod]
+        public async Task SupportAllOtherCurrentLanguages()
+        {
+            TestData.supportedTextandWordsTempInitializer();
+            Dictionary<string, string[]> supportedLangTextandWords = TestData.supportedTextandWords;
+            foreach (string key in supportedLangTextandWords.Keys)
```
bleroy: You should rename that variable to "language".

Tests/CustomEntitySearchTest/TestData.cs
```diff
@@ -39,7 +41,7 @@ public class TestData
     ]
 }";
         public const string emptyTextWordsNotFoundInput = @"""will you search?""";
-        public const string emptyWordsEmptyEntitiesInput = @"""if you find when searching, i will be sad""";
+        public const string emptyWordsEmptyEntitiesInput = @"""in""";
```
bleroy: I've just pushed a big refactoring of the test project, sorry, the merge will probably be tricky.

Tests/CustomEntitySearchTest/TestData.cs
```diff
@@ -267,5 +269,183 @@ public static async Task<WebApiSkillResponse> GeneratePayloadRequest(String inpu
                 request, new LoggerFactory().CreateLogger("local"), new Microsoft.Azure.WebJobs.ExecutionContext()));
             return (WebApiSkillResponse)response.Value;
         }
+
+        public static void supportedTextandWordsTempInitializer()
+        {
+            supportedTextandWords.Add("Greek", new string[]
+                {
+                    @"""Tου Αντώνη Ρέλλα - Θα πρέπει να γίνει κατανοητό ότι οι αποκλεισμοί των αναπήρων εκκινούν, έτσι κι αλλιώς, από τις θεσμοθετημένες πρακτικές του κράτους και τα εμπόδια στο δομημένο περιβάλλον. Πώς, λοιπόν, η κυβέρνηση θα κάνει πράξη την ανεξάρτητη διαβίωση""",
```
bleroy: Why do these strings have quotes inside them?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -50,38 +52,38 @@ public static class CustomEntitySearch
             }
 
             WebApiSkillResponse response = WebApiSkillHelpers.ProcessRequestRecords(skillName, requestRecords,
-                (inRecord, outRecord) => {
+                 (inRecord, outRecord) => {
                     string text = inRecord.Data["text"] as string;
                     IList<string> words = ((JArray)inRecord.Data["words"]).ToObject<List<string>>();
-                    if (words == null)
+                    Dictionary<string, string[]> synonyms = (inRecord.Data.ContainsKey("synonyms")) ? ((JContainer)inRecord.Data["synonyms"]).ToObject<Dictionary<string, string[]>>() : new Dictionary<string, string[]>();
```
bleroy: Can you break down long lines?

Tests/CustomEntitySearchTest/TestData.cs
bleroy: And you'll probably want to apply the same patterns from the refactored version: I've changed the code and sample data so it doesn't require so many deserializations, and used .NET types instead on serialized strings where possible.

Tests/CustomEntitySearchTest/TestData.cs
kargosh123: Prior to the .NET version I was using serialized strings. There should not be any quotes aside from the beginning and ending quotes.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -93,5 +95,146 @@ public static class CustomEntitySearch
...
+                                currWordChar = wordCharArray.Count - 1;
+                            }
+                        }
+                    }
+                    else
+                    {
+                        // fuzzy situation?
+                        if (currMismatch < leniency)
+                        {
+                            string accents = @"[\p{Mn}\p{Mc}]";
```
bleroy: You might want to make that a const, and maybe even a static readonly regexp

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -93,5 +95,146 @@ public static class CustomEntitySearch
...
+                            currMismatch += Regex.IsMatch(wordCharArray[currWordChar].ToString(), accents) ? 0.5 : 1;
```
bleroy: Also, a comment here would help understand what that regular expression is doing. The Unicode stuff is rather uncommon, and knowledge of how Unicode encodes accented characters is not that widespread. You might want to link to the Unicode spec or some similar resource. It's also very unclear how that's supposed to work. The way I read it, it seems to say that if a character is a mismatch, we only count it as half a mismatch if it is a diacritic/accented character? But that means that a 'ñ' for instance would count as half a mismatch even if the character in the word is 'e', which is completely unrelated.

Also, are you sure this is true of all accented characters? Take 'é': `Regex.IsMatch('é'.ToString(), @"[\p{Mn}\p{Mc}]")` is false.

Text/CustomEntitySearch/WordLinker.cs
```diff
@@ -10,14 +11,37 @@ internal class WordLinker
     {
         public WordLinker(string executingDirectoryPath)
         {
-            string json = File.ReadAllText($"{executingDirectoryPath}\\words.json");
-            Words = new List<string>(
-                JsonConvert.DeserializeObject<List<string>>(json));
+            string json = File.ReadAllText($"{executingDirectoryPath}\\Users\\t-neja\\azure-search-power-skills\\Text\\CustomEntitySearch\\words.json");
```
bleroy: What is this path?

Text/CustomEntitySearch/CustomEntitySearch.cs
jadrefke: >inRecord.Data["synonyms"] [](start = 115, length = 25)

Can you add model validations to the begining of this method so that you return "nice" erorr messages if the payload is malformed?

eg: right now if you supply malformed dictionaries you'll probably get back some hard to decipher json exception

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -93,5 +95,146 @@ public static class CustomEntitySearch
...
+            MatchCollection entityMatch = Regex.Matches(text, pattern, RegexOptions.IgnoreCase, TimeSpan.FromSeconds(MaxRegexEvalTime));
+            if (entityMatch.Count != 0)
+            {
+                foreach (Match match in entityMatch)
+                {
+                    entities.Add(
+                        new Entity
+                        {
+                            Category = "customEntity",
+                            Value = match.Value,
```
jadrefke: whats the purpose of this field if its always hardcoded to customEntity?

Text/CustomEntitySearch/WordLinker.cs
```diff
@@ -10,14 +11,37 @@ internal class WordLinker
...
+            get; private set;
+        }
+
+        private class Value
+        {
+            public IList<string> Words { get; set; }
+            public Dictionary<string, string[]> Synonyms { get; set; }
+            public IList<string> ExactMatch { get; set; }
+            public int FuzzyMatchOffset { get; set; }
+        }
```
jadrefke: why do you have a private class here? this looks identical to the parent class

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -50,38 +52,38 @@ public static class CustomEntitySearch
...
-                (inRecord, outRecord) => {
+                 (inRecord, outRecord) => {
                     string text = inRecord.Data["text"] as string;
                     IList<string> words = ((JArray)inRecord.Data["words"]).ToObject<List<string>>();
-                    if (words == null)
+                    Dictionary<string, string[]> synonyms = (inRecord.Data.ContainsKey("synonyms")) ? ((JContainer)inRecord.Data["synonyms"]).ToObject<Dictionary<string, string[]>>() : new Dictionary<string, string[]>();
+                    IList<string> exactMatches = (inRecord.Data.ContainsKey("exactMatches")) ? ((JArray)inRecord.Data["exactMatches"]).ToObject<List<string>>() : new List<string>();
+                    int offset = (inRecord.Data.ContainsKey("fuzzyMatchOffset") && (int)inRecord.Data["fuzzyMatchOffset"] >= 0) ? (int)inRecord.Data["fuzzyMatchOffset"] : 0;
+                    if (words.Count == 1 && words[0] == "")
                     {
```
jadrefke: This check looks like "if there is only one word thats an empty string, use json config based config"

Shouldn't it be if there's no words? Also validate that the other possible input fields aren't populated as well. You can throw an error and return a 400 if there's ambiguity of intent

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -50,38 +52,38 @@ public static class CustomEntitySearch
...
-                        words = new WordLinker(executionContext.FunctionAppDirectory).Words;
+                        WordLinker userInput = new WordLinker(executionContext.FunctionAppDirectory);
```
jadrefke: >WordLinker [](start = 51, length = 10)

Can you use this same model to deserialize the request payload too?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -93,5 +95,146 @@ public static class CustomEntitySearch
...
+                }
+                addWord = true;
+            }
+
+            if (leniency > 0)
+            {
+                // using KMP for now
+                // Create Table
+                IList<int> KMPTable = new List<int>(new int[checkMatch.Length + 1]);
+                IList<char> wordCharArray = checkMatch.ToCharArray();
```
jadrefke: imo this is algorithmically complicated enough to warrant its own unit tests. This is easiest to do when you refactor out the algorithmically significant code to its own class/helpers.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -93,5 +95,146 @@ public static class CustomEntitySearch
...
+                        }
+                    }
+                    else
+                    {
+                        // fuzzy situation?
+                        if (currMismatch < leniency)
+                        {
+                            string accents = @"[\p{Mn}\p{Mc}]";
+                            currMismatch += Regex.IsMatch(wordCharArray[currWordChar].ToString(), accents) ? 0.5 : 1;
+                            offset++;
```
jadrefke: check out [GetUnicodeCategory](https://docs.microsoft.com/en-us/dotnet/api/system.char.getunicodecategory?view=netframework-4.8) for an easier-to-read implementation of this

bleroy: And to address whether this is actually doing what it should be doing, take a look at this SO: https://stackoverflow.com/questions/359827/ignoring-accented-letters-in-string-comparison

Text/CustomEntitySearch/WordLinker.cs
kargosh123: Sorry, with my version of testing, I needed to use this path. Will delete.

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: The purpose of this field is to provide the "same" json output as NER skill. As there is no categorization in my code, I hardcoded it to customEntity

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: Which payload? Currently, the request payload can not have the keys and the skill will still run using the words.json file if no words are provided by the request payload. If the words.json file does not have any words, synonyms, etc. then the outRecord will just be empty.

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: Yes, but based on the existing method of using WebApiSkillResponse, I chose not to.

bleroy:
It looks like you need to rebase this: there are changes that appear in the PR that don't look to be actual.

Text/CustomEntitySearch/CustomEntitySearch.cs
jadrefke: do we need the same json schema across these two skills? I understand using it as a starting point to keep it similar, but if category doesn't provide value in this skill there's no reason to include it

---
In reply to: [312108230](https://github.com/Azure-Samples/azure-search-power-skills/pull/5#discussion_r312108230) [](ancestors = 312108230)

kargosh123:
> It looks like you need to rebase this: there are changes that appear in the PR that don't look to be actual.

I did rebase to integrate with your current setup?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+                IList<int> KMPTable = (caseSensitive) ? CreateKMPTable(checkMatch) : CreateKMPTable(checkMatch.ToLower());
+
+                // Begin searching!
+                int currTextChar = 0;
+                int currWordChar = 0;
+                StringWriter wordFound = new StringWriter();
+                double currMismatch = 0;
+                IList<char> wordCharArray = (caseSensitive) ? checkMatch.ToCharArray() : checkMatch.ToLower().ToCharArray();
+                IList<char> textCharArray = (caseSensitive) ? text.ToCharArray() : text.ToLower().ToCharArray();
+                IList<string> valuesList = new List<string>();
```
jadrefke: nit: you should be able to do this outside the for each word loop just to save a little redundant work

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+            }
+
+            if (leniency > 0)
+            {
+                // using KMP for now
+                // Create Table
+                IList<int> KMPTable = (caseSensitive) ? CreateKMPTable(checkMatch) : CreateKMPTable(checkMatch.ToLower());
+
+                // Begin searching!
+                int currTextChar = 0;
```
jadrefke: >currTextChar [](start = 20, length = 12)

nit: this looks like an index not a char, consider renaming "currTextCharIndex"

same for word char below

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+                int currWordChar = 0;
+                StringWriter wordFound = new StringWriter();
+                double currMismatch = 0;
+                IList<char> wordCharArray = (caseSensitive) ? checkMatch.ToCharArray() : checkMatch.ToLower().ToCharArray();
+                IList<char> textCharArray = (caseSensitive) ? text.ToCharArray() : text.ToLower().ToCharArray();
+                IList<string> valuesList = new List<string>();
+                IList<int> offsetList = new List<int>();
+                IList<double> confidenceList = new List<double>();
```
jadrefke: Could you just hve one list of entites here rather than 3 different lists that you have to stitch back together?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+                                Category = "customEntity",
+                                Value = valuesList[i],
+                                Offset = offsetList[i],
+                                Confidence = 1 - confidenceList[i]
+                            });
+                    }
+                }
+            }
+            if (addWord)
+                entitiesFound.Add(checkMatch);
```
jadrefke: why do you have this boolean? can you just call entitiesFound.Add(checkMatch); whenever you call entities.Add(~)?

kargosh123: No since depending on fuzzy/exact/multiple entities in the text, i only add entitiesFound once

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+                            wordFound.Write(text.ElementAt(currTextChar));
+                            currTextChar++;
+                            currWordChar++;
+
+                            if (currWordChar >= wordCharArray.Count || (currTextChar >= textCharArray.Count 
+                                && currMismatch + (wordCharArray.Count - currWordChar) <= leniency))
                             {
-                                foreach (Match match in entityMatch)
+                                // Code Cleanup?
+                                if (currMismatch > 0 || entityMatch.Count == 0)
```
jadrefke: >currMismatch > 0 [](start = 36, length = 16)

weird check, what does currMismatch < 0 represent?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
-                    outRecord.Data["Entities"] = entities;
-                    outRecord.Data["EntitiesFound"] = entitiesFound;
-                    return outRecord;
-                });
-
-            return new OkObjectResult(response);
+                for (int i = 0; i < valuesList.Count; i++)
+                {
+                    if (valuesList[i].Length == minVal)
+                    {
```
jadrefke: Couldn't you get multiple matches with variable length?

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: currMismatch = 0 is an exact match. sometimes, the regex matching doesn't find the exact match, so I had to add that case. currMismatch < 0 should not happen

jadrefke: It looks like this is always true?

---
In reply to: [313017607](https://github.com/Azure-Samples/azure-search-power-skills/pull/5#discussion_r313017607) [](ancestors = 313017607)

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -29,10 +30,10 @@ namespace AzureCognitiveSearch.PowerSkills.Text.CustomEntitySearch
     /// </summary>
     public static class CustomEntitySearch
     {
-        private static readonly int MaxRegexEvalTime = 1;
+        private static int MaxRegexEvalTime = 1;
         /// <summary>
         /// We assert the following assumptions:
-        /// 1. All text files are Latin based (i.e. no languageCode)
+        /// 1. All text files only characters provided in Unicode Character Database
```
jadrefke: >Database [](start = 76, length = 8)

nit: this is confusing, unicode is an encoding standard not a database

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+                         Dictionary<string, string[]> synonyms = (inRecord.Data.ContainsKey("synonyms")) ? 
+                         ((JContainer)inRecord.Data["synonyms"]).ToObject<Dictionary<string, string[]>>() : new Dictionary<string, string[]>();
+                         IList<string> exactMatches = (inRecord.Data.ContainsKey("exactMatches")) ? 
+                         ((JArray)inRecord.Data["exactMatches"]).ToObject<List<string>>() : new List<string>();
+                         int offset = (inRecord.Data.ContainsKey("fuzzyMatchOffset") && (int)(long)inRecord.Data["fuzzyMatchOffset"] >= 0) ? 
+                         (int)(long)inRecord.Data["fuzzyMatchOffset"] : 0;
+                         bool caseSensitive = (inRecord.Data.ContainsKey("caseSensitive")) ? (bool)inRecord.Data.ContainsKey("caseSensitive") : false;
+                         if (words.Count == 0 || (words.Count == 1 && words[0] == ""))
+                         {
+                             outRecord.Warnings.Add(new WebApiErrorWarningContract { Message = "The given key 'words' was not present in the dictionary." });
```
jadrefke: >The given key 'words' was not present in the dictionary. [](start = 96, length = 56)

nit: this warning doesn't help the user understand what the implications of not having words parameter present is. Maybe like "Used predefined key words from customLookupSkill configuration file since no 'words' parameter was supplied in web request"

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+                            else
+                            {
+                                currMismatch += 1;
+                                wordFound.Write(text.ElementAt(currTextChar));
+                                currTextChar++;
+                                currWordChar++;
                             }
 
+                            if((currWordChar >= wordCharArray.Count || (currTextChar >= textCharArray.Count
+                                && currMismatch + (wordCharArray.Count - currWordChar) <= leniency))  && currMismatch <= leniency)
```
jadrefke: >&& currMismatch <= leniency [](start = 102, length = 27)

isn't thi covered by the expression to the left of it?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+                            bool textHalfMark = Regex.Match(textCharArray[currTextChar].ToString(), accents).Success;
+
+                            // fuzzy situation?
+                            // accent case adds 0.5
+                            if ((wordHalfMark && textHalfMark) ||
+                                String.Compare(wordCharArray[currWordChar].ToString(), textCharArray[currTextChar].ToString(),
+                                CultureInfo.CurrentCulture, CompareOptions.IgnoreCase | CompareOptions.IgnoreNonSpace) == 0)
+                            {
+                                currMismatch += 0.5;
+                                wordFound.Write(text.ElementAt(currTextChar));
```
jadrefke: why mismatch += 0.5 if the strings match?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+                                currTextChar++;
+                                currWordChar++;
+                            }
+                            else if (currTextChar < textCharArray.Count - 1 && String.Compare(wordCharArray[currWordChar].ToString(), 
+                                textCharArray[currTextChar+1].ToString(), CultureInfo.CurrentCulture, CompareOptions.IgnoreCase | CompareOptions.IgnoreNonSpace) == 0)
```
jadrefke: >textCharArray[currTextChar+1].ToString() [](start = 32, length = 40)

this seems like a hardcoded lookahead. What happens if there's more than one modifying character?

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: They have different accents or one has an accent and the other does not

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: yes sorry!

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: the issue with checking case sensitivity outside of the for each word loop is that I cannot populate entities with the word found in the original text

bleroy: You should always be able to use a case-insentitive version of whatever API you use to do string or chard comparisons. In principle, you should not be using `ToLower` to achieve case insensitivity.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
-                    if (words == null)
+                 (inRecord, outRecord) => {
+                     if (!inRecord.Data.ContainsKey("text"))
+                     {
+                         outRecord.Errors.Add(new WebApiErrorWarningContract { Message = "The given key 'text' was not present in the dictionary." });
+                     }
+                     else if (!inRecord.Data.ContainsKey("words") && 
+                     (inRecord.Data.ContainsKey("synonyms") || inRecord.Data.ContainsKey("exactMatches") || inRecord.Data.ContainsKey("fuzzyMatchOffset")))
+                     {
+                         outRecord.Errors.Add(new WebApiErrorWarningContract { Message = "Cannot process record without the given key 'words' in the dictionary"});
```
bleroy: The error message doesn't match the actual test.

bleroy: nit: to slightly reduce the level of nesting, you can directly `return outRecord` from both those error cases, and do away with the final `else`.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+                         IList<string> words = (inRecord.Data.ContainsKey("words")) ? 
+                         ((JArray)inRecord.Data["words"]).ToObject<List<string>>() : new List<string>();
+                         Dictionary<string, string[]> synonyms = (inRecord.Data.ContainsKey("synonyms")) ? 
+                         ((JContainer)inRecord.Data["synonyms"]).ToObject<Dictionary<string, string[]>>() : new Dictionary<string, string[]>();
+                         IList<string> exactMatches = (inRecord.Data.ContainsKey("exactMatches")) ? 
+                         ((JArray)inRecord.Data["exactMatches"]).ToObject<List<string>>() : new List<string>();
+                         int offset = (inRecord.Data.ContainsKey("fuzzyMatchOffset") && (int)(long)inRecord.Data["fuzzyMatchOffset"] >= 0) ? 
+                         (int)(long)inRecord.Data["fuzzyMatchOffset"] : 0;
+                         bool caseSensitive = (inRecord.Data.ContainsKey("caseSensitive")) ? (bool)inRecord.Data.ContainsKey("caseSensitive") : false;
+                         if (words.Count == 0 || (words.Count == 1 && words[0] == ""))
```
bleroy: It looks a little arbitrary to look for `[""]`. How about `["", ""]`, etc.? Instead, you could add a `Where(w => !string.IsNullOrWhitespace(w))` earlier when building the list.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+                         var entities = new List<Entity>();
+                         var entitiesFound = new HashSet<string>();
+                         if (!string.IsNullOrWhiteSpace(text))
+                         {
+                             foreach (string word in words)
+                             {
+                                 if (string.IsNullOrEmpty(word)) continue;
+                                 int leniency = (exactMatches != null && exactMatches.Contains(word)) ? 0 : offset;
+                                 AddValues(word, text, entities, entitiesFound, leniency, caseSensitive);
+                                 if (synonyms != null && synonyms.ContainsKey(word))
```
bleroy: You can improve this a bit with `synonyms.TryGetValue(word, out string[] wordSynonyms)`.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -52,48 +53,273 @@ public static class CustomEntitySearch
...
+                        {
+                            string accents = @"[\p{Mn}\p{Sk}\p{Mc}]";
+                            bool wordHalfMark = Regex.Match(wordCharArray[currWordChar].ToString(), accents).Success;
+                            bool textHalfMark = Regex.Match(textCharArray[currTextChar].ToString(), accents).Success;
+
+                            // fuzzy situation?
+                            // accent case adds 0.5
+                            if ((wordHalfMark && textHalfMark) ||
+                                String.Compare(wordCharArray[currWordChar].ToString(), textCharArray[currTextChar].ToString(),
+                                CultureInfo.CurrentCulture, CompareOptions.IgnoreCase | CompareOptions.IgnoreNonSpace) == 0)
```
bleroy: Do you really need to do both the regex tests and the culture-aware string comparison?

Looks like the regex case is too wide, and will catch any pair of non-matching accented characters.

Are you sure the current culture should be used here? We know what language the text is in, so we should be able to infer a culture from it and use that.

Text/CustomEntitySearch/WordLinker.cs
```diff
@@ -14,13 +15,33 @@ internal class WordLinker
...
+            CaseSensitive = convert.CaseSensitive;
         }
 
         public IList<string> Words
         {
             get; private set;
         }
+        public IList<string> ExactMatch
+        {
+            get; private set;
```
bleroy: No need to specify `private set`: C# allows you to set get-only properties from your constructor. You only need this if you set those properties from anywhere else.

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: The issue currently is I do not ask for the language and the current culture is always english (and hence unnecessary will delete). Both are needed as in some instances the accent is considered nonspacing but still a different unicode value.

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: Is it better to have a warning or none at all?

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: Yes!

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: Also, as seen w. the test, it does not equate for instance a & e

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: Sorry, I'm not as familiar. Would the non-ToLower comparison repeat code but just ignoring case? Just confused

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: Thank you!

Text/CustomEntitySearch/CustomEntitySearch.cs
bleroy: You can't assume current culture is English. Since the skill deals with aspects of language that are not found in English, it may be reasonable actually to add language as an optional parameter of the skill. It's not a big constraint as language is already available from most skillsets.

On both being needed, reversely, are there cases that the string comparison would not find that the regexp would?

Text/CustomEntitySearch/CustomEntitySearch.cs
bleroy: What I'm saying is that all string comparison APIs in .NET have an overload that allows for case insensitivity, dictionaries can be configured to be case-insensitive, etc. `ToLower` should only really be used when you're going to actually use the lowercase version. If you make sure you use those APIs, then you don't need to modify the strings and make them lower, and it should make it safe to move the code out of the loop.

Text/CustomEntitySearch/CustomEntitySearch.cs
bleroy: I think none at all but you could get a second opinion on this.

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: Ah, I'm sorry. I meant that whenever I ran my tests with different languages, the culture remained english since its based on my UI (I think?). I am now just using the char class checks as that's all im really using the regexp for

Text/CustomEntitySearch/CustomEntitySearch.cs
bleroy: When we have the accent sensitivity flag, this should become 0 or 1.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -1,7 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.  
-// Licensed under the MIT License. See LICENSE file in the project root for full license information.  
-
```
jadrefke: dont remove copyright

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +128,310 @@ public static class CustomEntitySearch
...
+                double bestMismatchPreWhitespace = -1;
+                int prevWhiteSpaceIndex = 0;
+                int currTextCharIndex = 0;
+                int currWordCharIndex = 0;
+                StringWriter wordFound = new StringWriter();
+                double currMismatch = 0;
+                string wordCharArray = (caseSensitive) ? CreateWordArray(checkMatch) : CreateWordArray(checkMatch.ToLower(CultureInfo.CurrentCulture));
+                string textCharArray = (caseSensitive) ? text : text.ToLower(CultureInfo.CurrentCulture);
+                if (leniency >= wordCharArray.Length)
+                    leniency = wordCharArray.Length - 2;
```
jadrefke: co-locate this with the warning describing it

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +128,310 @@ public static class CustomEntitySearch
...
+
+                if (leniency >= 0)
+                {
+                    while (currTextCharIndex < textCharArray.Length)
+                    {
+                        // First find the delineating character for prefix addition later on
+                        if ((currWordCharIndex == 0 || currMismatch > 0) && textCharArray[currTextCharIndex].IsDelineating())
```
jadrefke: > textCharArray[currTextCharIndex].IsDelineating() [](start = 75, length = 49)

nit: you can combine this conditional and the one below it into one dealing with delineating characters

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +128,310 @@ public static class CustomEntitySearch
...
+                            prevWhiteSpaceIndex = currTextCharIndex;
+                        // Skip past extra delineating characters in the front of the word in the text
+                        if (currWordCharIndex == 0 && currMismatch >= 0 && textCharArray[currTextCharIndex].IsDelineating())
```
jadrefke: >currMismatch >= 0  [](start = 54, length = 18)

is this ever not true? i thought mismatch was always positive

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +128,310 @@ public static class CustomEntitySearch
...
+                            {
+                                if (bestMismatchPreWhitespace > -1 && bestMismatchPreWhitespace < currMismatch)
+                                {
+                                    entities.Add(
+                                    new Entity
+                                    {
+                                        Category = "customEntity",
+                                        Value = wordFound.GetStringBuilder().Remove(whitespaceOffset - initialOffsetIndex - 1, (int)bestMismatchPreWhitespace).ToString(),
+                                        Offset = initialOffsetIndex,
+                                        Confidence = leniency - bestMismatchPreWhitespace - (wordCharArray.Length - currWordCharIndex)
```
jadrefke: > leniency - bestMismatchPreWhitespace - (wordCharArray.Length - currWordCharIndex) [](start = 52, length = 82)

what does this number convey?

if i get an exact match, it would return the leniency? Should this just be the mismatch value?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +128,310 @@ public static class CustomEntitySearch
...
+                                    });
+
+                                    if (bestMismatchPreWhitespace > -1 && whitespaceOffset != initialOffsetIndex)
+                                    {
+                                        currTextCharIndex = whitespaceOffset;
+                                    }
+                                }
+
+                                if (!entitiesFound.Contains(checkMatch))
+                                    entitiesFound.Add(checkMatch);
```
jadrefke: nit: unnecessary check. Hashset.Add just turns false if its already there

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +128,310 @@ public static class CustomEntitySearch
...
+                                currWordCharIndex = 0;
+                                wordFound.GetStringBuilder().Clear();
+                                currMismatch = 0;
+                                bestMismatchPreWhitespace = -1;
+                            }
+                            else
+                                currMismatch = leniency + 1;
```
jadrefke: do you need this else statement? it looks like leniency is already > currmismatch to hit this?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +128,310 @@ public static class CustomEntitySearch
...
+
+                                }
+                            }
+                        }
+                        else
+                        {
+                            int initialOffsetIndex = currTextCharIndex - wordFound.ToString().Length;
+                            currWordCharIndex = 0;
+
+                            if (bestMismatchPreWhitespace > -1 && whitespaceOffset != initialOffsetIndex)
```
jadrefke: >(bestMismatchPreWhitespace > -1 && [](start = 31, length = 34)

why check that there is a bestMismatchPreWhitespace in this case? Regardless of if a reasonable match was found, you should still proceed to the next whitespace?

same for line ~380

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +128,310 @@ public static class CustomEntitySearch
...
+                    entitiesFound.Add(checkMatch);
+                }
+            }
+            else
+            {
+                // Begin searching!
+                int whitespaceOffset = 0;
+                double bestMismatchPreWhitespace = -1;
+                int prevWhiteSpaceIndex = 0;
+                int currTextCharIndex = 0;
```
jadrefke: this variable is unused?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +128,310 @@ public static class CustomEntitySearch
...
+                                    currWordCharIndex++;
+                                }
+                                else if (potWordMismatch > 0 && potTextMismatch == 0)
+                                {
+                                    currMismatch += 0.5;
+                                    currWordCharIndex++;
+                                }
+                                else if (potWordMismatch == 0 && potTextMismatch > 0)
+                                {
+                                    currMismatch += 0.5;
```
jadrefke: >0.5 [](start = 52, length = 3)

you're not using potWordMismatch as a double for these conditionals since you're just checking true/false and hardcoding currMismatch offset.

Can you change these conditionals to check textIsAccent and wordIsAccent booleans instead of this ternary double assignment?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +128,310 @@ public static class CustomEntitySearch
...
+                                                break;
+                                            }
+                                        }
+                                    }
+
+                                    if (trueTextComp && trueWordComp)
+                                    {
+                                        if (potWordMismatch == potTextMismatch)
+                                        {
+                                            currMismatch += potWordMismatch;
```
jadrefke: >potWordMismatch [](start = 60, length = 15)

does this handle the case:
Text: axbcz
Word: azbcx

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +128,310 @@ public static class CustomEntitySearch
...
+                                {
+                                    currMismatch += 0.5;
+                                    wordFound.Write(text[currTextCharIndex]);
+                                    currTextCharIndex++;
+                                }
+                                else
+                                {
+                                    int offsetWord = 0;
+                                    bool trueWordComp = false;
+                                    bool trueTextComp = false;
```
jadrefke: these variable names are kind of confusing, these look like "foundMatchInWordLookAhead" and "foundMatchInTextLookAhead"?

Text/CustomEntitySearch/CustomEntitySearch.cs
jadrefke: also, is it possible to consolidate this logic with the diffing logic below? The only difference is that accents are a different point value?

---
In reply to: [319582775](https://github.com/Azure-Samples/azure-search-power-skills/pull/5#discussion_r319582775) [](ancestors = 319582775)

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: Cannot co-locate because I would then be changing the entitiesFound word

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: This check was made for the "hello hello hello" text and "hello hello" word case

Tests/PowerSkillTests/Helpers.cs
```diff
@@ -42,6 +42,7 @@ public static async Task<WebApiSkillResponse> QueryFunction(string inputText, Fu
                 Method = "POST"
             };
             var response = (OkObjectResult)(await function(request));
+            Console.WriteLine(JsonConvert.SerializeObject((WebApiSkillResponse)response.Value));
```
bleroy: Oops, looks like leftover debug code.

bleroy: By the way, when I do this sort of console debugging, I usually use `System.Diagnostics.Debug.WriteLine` instead of `Console`.

Text/CustomEntitySearch/PotentialEntity.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace AzureCognitiveSearch.PowerSkills.Text.CustomEntitySearch
+{
+    public class PotentialEntity
```
jadrefke: >PotentialEntity [](start = 17, length = 15)

file looks unused in most recent iteration

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +132,174 @@ public static class CustomEntitySearch
...
+                {
+                    if (textCharArray[currTextCharIndex].IsDelineating())
+                        pointersInText.Add(currTextCharIndex + 1);
+                }
+                pointersInText.Add(textCharArray.Length + 1);
+
+                double[] minLevenshteinDistance = new double[pointersInText.Count];
+                int[] endofMatchInTextPointer = new int[pointersInText.Count];
+                if (leniency > wordCharArray.Length - 1)
+                    leniency = Math.Max(0, wordCharArray.Length - 1);
```
jadrefke: can you move this math up to where you're adding the warning to the response so the two are co-located

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -50,40 +56,71 @@ public static class CustomEntitySearch
...
+                    int offset = (inRecord.Data.ContainsKey("fuzzyMatchOffset")) ? Math.Max(0, Convert.ToInt32(inRecord.Data["fuzzyMatchOffset"])) : 0;
+                    bool caseSensitive = (inRecord.Data.ContainsKey("caseSensitive")) ? (bool)inRecord.Data.ContainsKey("caseSensitive") : false;
+                    if (words.Count == 0 || (words.Count(word => !String.IsNullOrEmpty(word)) == 0))
                     {
-                        words = new WordLinker(executionContext.FunctionAppDirectory).Words;
+                        outRecord.Warnings.Add(new WebApiErrorWarningContract {
+                            Message = "Used predefined key words from customLookupSkill configuration file " +
+                            "since no 'words' parameter was supplied in web request" });
+                        WordLinker userInput = WordLinker.WordLink(executionContext.FunctionAppDirectory);
+                        words = userInput.Words;
```
jadrefke: can you try/catch this and provide a specific error message like "could not parse predefined words.json" instead of some potentially ambiguous "failed to deserialize" error

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +132,174 @@ public static class CustomEntitySearch
...
+            }
+            else
+            {
+                List<int> pointersInText = new List<int>{ 0 };
+                string textCharArray = (caseSensitive) ? text : text.ToLower(CultureInfo.CurrentCulture);
+                for (int currTextCharIndex = 0; currTextCharIndex < textCharArray.Length; currTextCharIndex++)
+                {
+                    if (textCharArray[currTextCharIndex].IsDelineating())
+                        pointersInText.Add(currTextCharIndex + 1);
+                }
```
jadrefke: this looks like it may add unnecessary pointers (eg: space at end of text, multiple spaces right next to each other)

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +132,174 @@ public static class CustomEntitySearch
...
+                pointersInText.Add(textCharArray.Length + 1);
+
+                double[] minLevenshteinDistance = new double[pointersInText.Count];
+                int[] endofMatchInTextPointer = new int[pointersInText.Count];
+                if (leniency > wordCharArray.Length - 1)
+                    leniency = Math.Max(0, wordCharArray.Length - 1);
+                for (int pointer = 0; pointer < pointersInText.Count - 1; pointer++)
```
jadrefke: >pointer [](start = 25, length = 7)

nit: this is index to the pointer, not the pointer itself. Mayube rename variable to make that clear?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +132,174 @@ public static class CustomEntitySearch
             return new OkObjectResult(response);
         }
 
+        public static void AddValues(string checkMatch, string text, List<Entity> entities, 
+            HashSet<string> entitiesFound, int leniency, bool caseSensitive)
```
jadrefke: nit: can you put each parameter on its own line

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +132,174 @@ public static class CustomEntitySearch
...
+                        substitutionCost = 0;
+                    else if (String.Compare(checkMatch[currWordIndex].ToString(), text[currTextIndex].ToString(),
+                                    CultureInfo.CurrentCulture, CompareOptions.IgnoreNonSpace) == 0)
+                        substitutionCost = .5;
+                    else
+                        substitutionCost = 1;
+                    if (checkMatch[currWordIndex].IsAccent() ^ text[currTextIndex].IsAccent())
+                        accentAddition = 0.5;
+                    dynamicDistanceCalc[currTextIndex + 1, currWordIndex + 1] = Math.Min(
+                        Math.Min(dynamicDistanceCalc[currTextIndex, currWordIndex + 1] + accentAddition,
```
jadrefke: >dynamicDistanceCalc[ [](start = 33, length = 20)

nit: can you put each of these statements on a seperate line and add comments for which is substitution/addition/removal

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -132,10 +149,15 @@ public static class CustomEntitySearch
             return new OkObjectResult(response);
         }
 
-        public static void AddValues(string checkMatch, string text, List<Entity> entities, 
-            HashSet<string> entitiesFound, int leniency, bool caseSensitive)
+        public static void AddValues(
+            string checkMatch, 
+            string text, 
+            string wordCharArray,
```
bleroy: Why is this still called `wordCharArray`? Should it be just `word`?

bleroy:
Can you rebase on the latest master? #9 in particular should require a couple changes.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +149,181 @@ public static class CustomEntitySearch
...
+                    if (textCharArray[currTextCharIndex].IsDelineating())
+                    {
+                        if (currTextCharIndex + 1 < textCharArray.Length && !textCharArray[currTextCharIndex + 1].IsDelineating())
+                            startPointersInText.Add(currTextCharIndex + 1);
+                        if (currTextCharIndex - 1 >= 0 && !textCharArray[currTextCharIndex - 1].IsDelineating())
+                            endPointersInText.Add(currTextCharIndex - 1);
+                    }
+                }
+                endPointersInText.Add(textCharArray.Length - 1);
+
```
jadrefke: this might be extra if the text ensd with a delineator?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +149,181 @@ public static class CustomEntitySearch
...
+                double[] minLevenshteinDistance = new double[startPointersInText.Count];
+                int[] endofMatchInTextPointer = new int[startPointersInText.Count];
+                for (int startIndex = 0; startIndex < startPointersInText.Count; startIndex++)
```
jadrefke: >startIndex [](start = 25, length = 10)

maybe startPointerIndex

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +149,181 @@ public static class CustomEntitySearch
...
+                {
+                    minLevenshteinDistance[startIndex] = -1;
+                    for (int endIndex = startIndex; endIndex < endPointersInText.Count; endIndex++)
+                    {
+                        double distance = DamerauLevenshteinCalculation(textCharArray.Substring(startPointersInText[startIndex],
+                            endPointersInText[endIndex] - startPointersInText[startIndex] + 1), wordCharArray);
```
jadrefke: >endPointersInText[endIndex] - startPointersInText[startIndex] + 1 [](start = 28, length = 65)

you can short circuit out before doing the calculation if the length of this string is too big

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +149,181 @@ public static class CustomEntitySearch
...
+                            endPointersInText.Add(currTextCharIndex - 1);
+                    }
+                }
+                endPointersInText.Add(textCharArray.Length - 1);
+
+                double[] minLevenshteinDistance = new double[startPointersInText.Count];
+                int[] endofMatchInTextPointer = new int[startPointersInText.Count];
+                for (int startIndex = 0; startIndex < startPointersInText.Count; startIndex++)
+                {
+                    minLevenshteinDistance[startIndex] = -1;
```
jadrefke: >-1 [](start = 57, length = 2)

could initialize these to leniancy + 1 and save a check on line 226

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +149,181 @@ public static class CustomEntitySearch
...
+                    for (int endIndex = startIndex; endIndex < endPointersInText.Count; endIndex++)
+                    {
+                        double distance = DamerauLevenshteinCalculation(textCharArray.Substring(startPointersInText[startIndex],
+                            endPointersInText[endIndex] - startPointersInText[startIndex] + 1), wordCharArray);
+                        if (distance > -1 && (minLevenshteinDistance[startIndex] == -1 || minLevenshteinDistance[startIndex] > distance))
+                        {
+                            minLevenshteinDistance[startIndex] = distance;
+                            endofMatchInTextPointer[startIndex] = endIndex;
+                        }
+                        else if (distance > checkMatch.Length * 2)
```
jadrefke: >checkMatch.Length * 2) [](start = 44, length = 22)

why *2?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +149,181 @@ public static class CustomEntitySearch
...
+                    if (checkMatch[currWordIndex].IsAccent() ^ text[currTextIndex].IsAccent())
+                        accentAddition = 0.5;
+                    dynamicDistanceCalc[currTextIndex + 1, currWordIndex + 1] = Math.Min(
+                        Math.Min(dynamicDistanceCalc[currTextIndex, currWordIndex + 1] + accentAddition, // deletion
+                        dynamicDistanceCalc[currTextIndex + 1, currWordIndex] + accentAddition), // insertion
+                        dynamicDistanceCalc[currTextIndex, currWordIndex] + substitutionCost); // substitution
+                    if (currTextIndex > 0 && currWordIndex > 0 && text[currTextIndex].Equals(checkMatch[currWordIndex - 1]) && checkMatch[currWordIndex].Equals(text[currTextIndex - 1]))
+                        dynamicDistanceCalc[currTextIndex + 1, currWordIndex + 1] = Math.Min(dynamicDistanceCalc[currTextIndex + 1, currWordIndex + 1], 
+                            dynamicDistanceCalc[currTextIndex - 1, currWordIndex - 1] + substitutionCost); // transposition
+                }
```
jadrefke: this looks like you're having like 

aaazxbbb
aaaxzbbb

count as a diff of 1? Am i reading that right?

Text/CustomEntitySearch/CustomEntitySearch.cs
kargosh123: just to reduce absurd checks

kargosh123: will change based on your previous feedback

Text/CustomEntitySearch/CustomEntitySearch.cs
jadrefke: synced offline, yes this is by design

---
In reply to: [322980495](https://github.com/Azure-Samples/azure-search-power-skills/pull/5#discussion_r322980495) [](ancestors = 322980495)

kargosh123:
> Can you rebase on the latest master? #9 in particular should require a couple changes.

Finished rebase!

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +166,180 @@ public static class CustomEntitySearch
...
+            {
+                // Overlap checker now also included in Regex expression using delineating characters as overlap lookahead
+                StringBuilder escapedWord = new StringBuilder(@"(?=(");
+                if (!word.First().IsDelineating() && !substringMatch)
+                    escapedWord.Append(@"\b");
+                for (int currWordCharIndex = 0; currWordCharIndex < word.Length; currWordCharIndex++)
+                {
+                    if (word[currWordCharIndex].IsDelineating())
+                    {
+                        escapedWord.Append(".");
```
jadrefke: >("."); [](start = 42, length = 6)

should this be \s?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +166,180 @@ public static class CustomEntitySearch
...
+                endPointersInText.Add(textCharArray.Length - 1);
+
+                double[] minLevenshteinDistance = new double[startPointersInText.Count];
+                int[] endofMatchInTextPointer = new int[startPointersInText.Count];
+                for (int startPointerIndex = 0; startPointerIndex < startPointersInText.Count; startPointerIndex++)
+                {
+                    minLevenshteinDistance[startPointerIndex] = leniency + 1;
+                    for (int endPointerIndex = startPointerIndex; endPointerIndex < endPointersInText.Count; endPointerIndex++)
+                    {
+                        if (endPointersInText[endPointerIndex] - startPointersInText[startPointerIndex] + 1 > checkMatch.Length * 2) break;
```
jadrefke: >endPointersInText[endPointerIndex] - startPointersInText[startPointerIndex] + 1 [](start = 28, length = 79)

nit: this expression is used twice. Might be nice to store in tmp variable for readability (eg: potentialMatchLength

jadrefke: >checkMatch.Length * 2 [](start = 110, length = 21)

wordLength + leniency would be a stricter bound

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -95,5 +166,180 @@ public static class CustomEntitySearch
...
+            }
+
+            return dynamicDistanceCalc[text.Length, checkMatch.Length];
+        }
+
+        /*
+        * Given an entity the user wants to find, this method removes delineating characters if they are found in the
+        * beginning or end of the entity definition. The method then returns the exact word that will be used for fuzzy matching
+        */
+        public static string CreateWordArray(string checkMatch)
```
jadrefke: >CreateWordArray [](start = 29, length = 15)

nit: rename to maybe "TrimDelineatingCharacters()" since you use it for text as well as words

bleroy:
This work has been merged as #10

PR #7, opened 2019-08-30
Test acronyms, Bing entity search, tokenizer, fix some metadata

bleroy:
## Purpose
Test acronyms, Bing entity search, tokenizer, fix some metadata

No

## Pull Request Type
Tests and minor fixes

### DIFF
--- a/Text/BingEntitySearch/BingEntity.cs
+++ b/Text/BingEntitySearch/BingEntity.cs
@@ -6,7 +6,7 @@ namespace AzureCognitiveSearch.PowerSkills.Text.BingEntitySearch
     public class BingEntity
     {
 
-        public Contractualrule[] contractualRules { get; set; }
+        public Contractualrule[] ContractualRules { get; set; }
         public Image Image { get; set; }
         public string Description { get; set; }
         public string BingId { get; set; }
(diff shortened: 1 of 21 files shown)


Text/BingEntitySearch/BingEntity.cs
```diff
@@ -6,7 +6,7 @@ namespace AzureCognitiveSearch.PowerSkills.Text.BingEntitySearch
     public class BingEntity
     {
 
-        public Contractualrule[] contractualRules { get; set; }
+        public Contractualrule[] ContractualRules { get; set; }
```
jadrefke: >r [](start = 26, length = 1)

nit: cammel case type

Text/Tokenizer/Tokenizer.cs
```diff
@@ -33,26 +33,51 @@ public static class Tokenizer
...
+                    outRecord.Data["words"] = predictionEngine.Predict(text).Words ?? Array.Empty<string>();
                     return outRecord;
                 });
 
             return new OkObjectResult(response);
         }
 
+        private static StopWordsRemovingEstimator.Language MapToMlNetLanguage(string languageCode)
+        {
+            switch(languageCode)
```
jadrefke: >languageCode [](start = 19, length = 12)

do you want to trim & tolower?

PR #8, opened 2019-09-03
Add a distinct skill to de-duplicate a list of terms

bleroy:
## Purpose

Add a skill the de-duplicates a list of terms

No

## Pull Request Type

Feature


Text/Distinct/Thesaurus.cs
```diff
@@ -0,0 +1,61 @@
...
+            var normalizedToWord = new Dictionary<string, string>();
+            foreach (string word in words)
+            {
+                string normalized = Normalize(word);
+                string canonical = Synonyms.TryGetValue(normalized, out string canonicalFromThesaurus) ?
+                    canonicalFromThesaurus :
+                    normalized;
+                if (!normalizedToWord.ContainsKey(canonical))
+                {
+                    normalizedToWord.Add(canonical, canonicalFromThesaurus ?? word); // Arbitrarily consider the first occurrence as canonical
```
jadrefke: >word [](start = 78, length = 4)

do you want to use the normalized form here?

would be kind of weird if the distinct word that came back was like "time...." because the first occurrence was "Once upon a time..." even though that word doesn't appear as such anywhere else in the list


If you do decide to go with that, you can probably make normalizedToWord just a hash set

Text/Distinct/Thesaurus.cs
```diff
@@ -0,0 +1,61 @@
...
+            : this(JsonConvert.DeserializeObject<IEnumerable<IEnumerable<string>>>(
+                File.ReadAllText($"{executingDirectoryPath}\\thesaurus.json")))
+        { }
+
+        public Thesaurus(IEnumerable<IEnumerable<string>> dataset)
+        {
+            Synonyms = new Dictionary<string, string>(StringComparer.InvariantCultureIgnoreCase);
+            foreach (IEnumerable<string> lemma in dataset)
+            {
+                string canonicalForm = lemma.First();
```
jadrefke: >.First(); [](start = 44, length = 9)

can gracefully handle empty arrays by just continuing here. 

First throws if an element is not present

Text/Distinct/Thesaurus.cs
```diff
@@ -0,0 +1,61 @@
...
+                foreach (string form in lemma)
+                {
+                    Synonyms.Add(Normalize(form), canonicalForm);
```
jadrefke: >Add [](start = 29, length = 3)

this throws if the element is already present - which might happen since you're normalizing the forms?

may want to check if it already exists, and if so is the value the same as this new canonicalForm - and only throw if they mismatch (ambiguity of intent)

Text/Distinct/Thesaurus.cs
```diff
@@ -0,0 +1,61 @@
...
+                    normalizedToWord.Add(canonical, canonicalFromThesaurus ?? word); // Arbitrarily consider the first occurrence as canonical
+                }
+            }
+            return normalizedToWord.Values.Distinct();
+        }
+
+        public static string Normalize(string word)
+            => new string(word
+                .Normalize()
+                .ToLower()
```
jadrefke: >ToLower [](start = 17, length = 7)

nit: ToLowerInvariant?

Text/Distinct/Distinct.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            string skillName = executionContext.FunctionName;
+            IEnumerable<WebApiRequestRecord> requestRecords = WebApiSkillHelpers.GetRequestRecords(req);
+            if (requestRecords == null)
+            {
+                return new BadRequestObjectResult($"{skillName} - Invalid request record array.");
+            }
+
+            Thesaurus thesaurus = new Thesaurus(executionContext.FunctionAppDirectory);
+            WebApiSkillResponse response = WebApiSkillHelpers.ProcessRequestRecords(skillName, requestRecords,
```
jadrefke: nit: try/catch with error message along the lines of "could not parse thesaurus json" instead of some potentially ambiguous "unable to deserialize" message

Text/Distinct/Thesaurus.cs
```diff
@@ -0,0 +1,61 @@
...
+                foreach (string form in lemma)
+                {
+                    Synonyms.Add(Normalize(form), canonicalForm);
+                }
+            }
+        }
+
+        public Dictionary<string, string> Synonyms
+        {
+            get; private set;
```
jadrefke: >private set; [](start = 17, length = 12)

nit: doesn't look like you need a private set - its only assigned in constructor

Text/Distinct/README.md
```diff
@@ -0,0 +1,100 @@
...
+- azure-search
+name: Distinct sample skill for cognitive search
+description: This custom skill removes duplicates from a list of terms.
+azureDeploy: https://raw.githubusercontent.com/Azure-Samples/azure-search-power-skills/master/Text/Distinct/azuredeploy.json
+---
+
+# Distinct
+
+This custom skill removes duplicates from a list of terms.
+
```
jadrefke: include that the skill is also case insensitive and ignores punctuation and seperator characters

Text/Distinct/Distinct.cs
```diff
@@ -0,0 +1,45 @@
...
+                (inRecord, outRecord) =>
+                {
+                    var words = ((JArray)inRecord.Data["words"]).Values<string>();
```
jadrefke: >((JArray)inRecord.Data["words"]).Values<string>(); [](start = 32, length = 50)

handle null with an explicit error/warning message as well?

Text/Distinct/Thesaurus.cs
bleroy: I had the same concern, and actually decided against it because the normalized form is actually pretty weird when spaces and punctuation are involved. The reasoning here is that the normalized form is considered an internal representation used for comparisons, but should not constitute output. So sure, using the first occurrence is arbitrary and can be weird, but that can be mitigated by adding an entry with an explicit canonical form to the thesaurus. I actually don't think it will be a big issue in the context of extracted entities, which was the main scenario motivating building this skill.

And yes, I implemented it first with a hash set, and it was wayyyy simpler ;)

PR #9, opened 2019-09-09
Adding support for CSV

liamca:
## Purpose

* ...

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -29,6 +29,9 @@ namespace AzureCognitiveSearch.PowerSkills.Text.CustomEntitySearch
     /// </summary>
     public static class CustomEntitySearch
     {
+        // Use this to load from "csv" or "json" file 
+        public static IList<string> preLoadedWords = new WordLinker("csv").Words;
```
bleroy: I'm fine with the default being CSV, but we should try to lookup an env variable first, and if present, use that. `WordLinker` could also check for file existence, and fallback gracefully.

bleroy:
Actually, we discussed it, and we'll merge this first and we'll rebase the other one later.

PR #14, opened 2019-10-01
Improve regex perf for CustomLookupSkill and refactor code for readability

jadrefke:
## Purpose

* ...

```
[ ] Yes
[ x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


Tests/PowerSkillTests/CustomEntitySearchTests/CustomEntitySearchHelpers.cs
```diff
@@ -76,14 +85,60 @@ public static string BuildOutput(string[] entities, string[] matches, int[] matc
...
+
+            JObject config = new JObject();
+            config["words"] = JArray.FromObject(words);
+            config["synonyms"] = JObject.FromObject(synonyms);
+            config["exactMatch"] = JArray.FromObject(exactMatches);
+            config["fuzzyEditDistance"] = offset;
+            config["caseSensitive"] = true;
+
+            var executingLocation = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            var path = Path.Combine(executingLocation, "testWords.json");
```
bleroy: Constants for this and the csv one would be good.

Tests/PowerSkillTests/Helpers.cs
```diff
@@ -96,9 +100,27 @@ public static HttpResponseMessage RespondRequestWith(this HttpRequestMessage req
         }
 
         public static Func<HttpRequest, Task<IActionResult>> CurrySkillFunction(Func<HttpRequest, ILogger, Microsoft.Azure.WebJobs.ExecutionContext, Task<IActionResult>> skillFunction)
-            => request => skillFunction(request, new LoggerFactory().CreateLogger("local"), new Microsoft.Azure.WebJobs.ExecutionContext());
+        {
+            var executionContext = new Microsoft.Azure.WebJobs.ExecutionContext();
+            var executingLocation = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
```
bleroy: We do this so much, we should just build a small helper for it. Not necessarily in this PR though.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -4,162 +4,142 @@
...
         {
-            if (preLoadedWords == null)
+            log.LogInformation("Custom Entity Search function: C# HTTP trigger function processed a request.");
+
+            if (_userDefinedEntities == null || _precompiledExactMatchRegex == null
+                || executionContext.FunctionName == "unitTestFunction") // always reload data for tests
             {
-                preLoadedWords = WordLinker.WordLink(executionContext.FunctionAppDirectory, "csv").Words;
+                _userDefinedEntities = WordLinker.WordLink(EntityDefinitionLocation);
+                _precompiledExactMatchRegex = new Regex($@"\b({string.Join("|", _userDefinedEntities.Words)})\b",
```
bleroy: It looks like it's possible words would contain regex special characters. They should probably be escaped first.

Tests/PowerSkillTests/CustomEntitySearchTests/CustomEntitySearchHelpers.cs
```diff
@@ -76,14 +85,60 @@ public static string BuildOutput(string[] entities, string[] matches, int[] matc
...
         {
-            string input = BuildInput(text, words);
+            string input = BuildInput(text);
+            ReplaceWordsJson(words, synonyms, exactMatches, offset);
+            string actualOutput = await QueryEntitySearchFunctionAndSerialize(input);
+
+            Helpers.AssertJsonEquals(expectedOutput, actualOutput);
+        }
+
+        public static void ReplaceWordsJson(
```
bleroy: Can you rename this to make it more explicit it's replacing the words file? Like maybe `ReplaceWordsJsonFile` or `UpdateWordsJsonFile`?

Tests/PowerSkillTests/CustomEntitySearchTests/TestData.cs
```diff
@@ -106,21 +105,11 @@ public static class TestData
...
..."":""simple, I"",""offset"":15,""confidence"":9.0},{""category"":""customEntity"",""value"":""I overlooked it"",""offset"":23,""confidence"":5.0},{""category"":""customEntity"",""value"":""overlooked it"",""offset"":25,""confidence"":3.0},{""category"":""customEntity"",""value"":""Should"",""offset"":42,""confidence"":8.0},{""category"":""customEntity"",""value"":""work on"",""offset"":49,""confidence"":8.0},{""category"":""customEntity"",""value"":""on all tests"",""offset"":54,""confidence"":9.0},{""category"":""customEntity"",""value"":""all"",""offset"":57,""confidence"":9.0},{""category"":""customEntity"",""value"":""tests now"",""offset"":61,""confidence"":9.0},{""category"":""customEntity"",""value"":""now"",""offset"":67,""confidence"":9.0}],""entitiesFound"":[""fix"",""soo "",""overlooking"",""overlooked""]},""errors"":[],""warnings"":[{""message"":""The provided fuzzy offset of 10, is larger than the length of the provided word, \""fix\"".""},{""message"":""The provided fuzzy offset of 10, is larger than the length of the provided word, \""soo \"".""},{""message"":""The provided fuzzy offset of 10, is larger than the length of the provided word, \""overlooked\"".""}]}]}";
```
bleroy: Serialized JSON is harder to read and maintain. Can we define this as an anonymous object and serialize it in code instead?

Tests/PowerSkillTests/CustomEntitySearchTests/TestData.cs
```diff
@@ -606,14 +595,6 @@ public static void supportedTextandWordsTempInitializer()
...
             supportedMatchIndices.Add("German", new int[] { 387 });
             supportedConfidence.Add("German", new double[] { 0.5 });
-            supportedTextandWords.Add("Basic Overlap", new string[]
-                {
-                    @"a a a c a a c",
-                    @"a a",
-                    @"a a, a a, a a",
-                    @"a a"
-                });
-            supportedMatchIndices.Add("Basic Overlap", new int[] { 0, 2, 8 });
```
bleroy: So why are we removing tests? Are those cases no longer relevant?

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -4,162 +4,142 @@
...
+    /// this function will find all occurences of that entity in some input text.
     /// </summary>
     public static class CustomEntitySearch
     {
-        // Use this to load from "csv" or "json" file 
-        public static IList<string> preLoadedWords = null;
+        // ** Some global variables used for configuration.**
+        // ** Change these values prior to deploying your function **
+        public static bool ExactMatchesShouldBeCaseSensitive = false;
+        public static string EntityDefinitionLocation = "words.csv"; // other option is "words.json"
```
bleroy: Not new from this PR, but eventually, we should lookup an environment variable and let it override this default instead of requiring rebuilding the code.

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -4,162 +4,142 @@
...
-                                     Message = @"The provided fuzzy offset of " + leniency + @", is larger than the length of the provided word, """ + word + @"""."
-                                 });
-                                 leniency = Math.Max(0, wordCharArray.Length - 1);
+                                     entities.Add(
+                                         new Entity
+                                         {
+                                             Category = "customEntity",
+                                             Value = match.Groups[1].Value,
+                                             Offset = match.Index,
+                                             Confidence = 0
```
bleroy: Seems weird to assign zero confidence to an exact match (but I'm assuming this is never used for exact matches?)

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -173,180 +153,207 @@ public static class CustomEntitySearch
...
+            {
+                return potentialMatch?.Length ?? -1;
+            }
+
+            double[,] dynamicDistanceCalc = new double[potentialMatch.Length + 1, entityToFind.Length + 1];
+
+            for (int currpotentialEntityMatchIndex = 0; currpotentialEntityMatchIndex <= potentialMatch.Length; currpotentialEntityMatchIndex++)
+                dynamicDistanceCalc[currpotentialEntityMatchIndex, 0] = currpotentialEntityMatchIndex;
+            for (int currWordIndex = 0; currWordIndex <= entityToFind.Length; currWordIndex++)
                 dynamicDistanceCalc[0, currWordIndex] = currWordIndex;
```
bleroy: nit: braces

Text/CustomEntitySearch/CustomEntitySearch.cs
```diff
@@ -173,180 +153,207 @@ public static class CustomEntitySearch
...
+
+            while (startIndex < textToTrim.Length && textToTrim[startIndex].IsDelineating())
+                startIndex++;
+            while (endIndex >= 0 && textToTrim[endIndex].IsDelineating())
+                endIndex--;
+
+            if (startIndex >= endIndex)
             {
-                return checkMatch.Substring(initCheckIndex, endCheckIndex - initCheckIndex + 1);
+                return string.Empty;
```
bleroy: nit: `""`

Text/CustomEntitySearch/WordLinker.cs
```diff
@@ -13,42 +13,60 @@ namespace AzureCognitiveSearch.PowerSkills.Text.CustomEntitySearch
...
-            else if (fileType == "csv")
+            else if (fileName.EndsWith(".csv"))
             {
                 return new WordLinker
                 {
-                    Words = File.ReadAllLines($"{executingPathDirectory}\\words.csv").ToList()
+                    Words = File.ReadAllLines(Path.Join(actual_root, fileName))
+                            .SelectMany(line => line.Split(","))
+                            .Where(line => !string.IsNullOrEmpty(line))
+                            .ToList()
```
bleroy: This is an oversimplification of CSV. We should either clarify in the readme that we only accept comma-separated, no quotes, or bring in one of the simple CSV packages that exist.

Tests/PowerSkillTests/CustomEntitySearchTests/TestData.cs
jadrefke: going to won't fix for this PR. I plan on taking another stab at refactoring tests in the next PR

---
In reply to: [330754632](https://github.com/Azure-Samples/azure-search-power-skills/pull/14#discussion_r330754632) [](ancestors = 330754632)

Tests/PowerSkillTests/CustomEntitySearchTests/TestData.cs
jadrefke: yes, i removed the code that lets you specify words in the request payload. You must use the config files now (similar to the schema we've defined for the 1st part integration)

---
In reply to: [330755115](https://github.com/Azure-Samples/azure-search-power-skills/pull/14#discussion_r330755115) [](ancestors = 330755115)

Text/CustomEntitySearch/CustomEntitySearch.cs
jadrefke: sooo the "confidence" is actually the edit distance.. so 0 means there's no characters different. I didn't want to mess with the schema too much since they're using this skill in workshops this week

---
In reply to: [330762711](https://github.com/Azure-Samples/azure-search-power-skills/pull/14#discussion_r330762711) [](ancestors = 330762711)

Text/CustomEntitySearch/CustomEntitySearch.cs
jadrefke: eh, i prefer string.Empty. Makes it explicit that i intended it to be empty. Let me know if you feel strongly about "" though

---
In reply to: [330769371](https://github.com/Azure-Samples/azure-search-power-skills/pull/14#discussion_r330769371) [](ancestors = 330769371)

Text/CustomEntitySearch/WordLinker.cs
jadrefke: we may need to take that change after workshops are over. Going to won't fix for the scope of this PR. 

---
In reply to: [330772512](https://github.com/Azure-Samples/azure-search-power-skills/pull/14#discussion_r330772512) [](ancestors = 330772512)

Text/CustomEntitySearch/CustomEntitySearch.cs
bleroy: Oooh, right. OK. Yes, that's confusing. Should be renamed eventually.

Text/CustomEntitySearch/CustomEntitySearch.cs
bleroy: I mildly feel `""` is always better, and a little more strongly we should be consistent throughout this codebase, but eh, such a small and inconsequential detail...

Text/CustomEntitySearch/WordLinker.cs
bleroy: I'd still add a comment in the readme and/or a bug to acknowledge we're aware of it.

PR #36, opened 2020-09-02
New Power Skill DecryptBlobFile

Careyjmac:
## Purpose
Introduces a new power skill that allows customers to index encrypted files stored in Azure Blob Storage.
* ...

```
[ ] Yes
[ x] No
```

## Pull Request Type

```
[ ] Bugfix
[x ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code
Follow the README guide to deploy the function and test the workflow.
```
```


Utils/DecryptBlobFile/DecryptBlobFile.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            WebApiSkillResponse response = WebApiSkillHelpers.ProcessRequestRecords(skillName, requestRecords,
+                (inRecord, outRecord) =>
+                {
+                    string blobUrl = (inRecord.Data.TryGetValue("blobUrl", out object blobUrlObject) ? blobUrlObject : null) as string;
+                    string sasToken = (inRecord.Data.TryGetValue("sasToken", out object sasTokenObject) ? sasTokenObject : null) as string;
+
+                    if (string.IsNullOrWhiteSpace(blobUrl))
+                    {
+                        outRecord.Errors.Add(new WebApiErrorWarningContract() { Message = "Parameter 'blobUrl' is required to be present and a valid uri." });
```
bleroy, line 53: nit: `$""` string with `{nameof(blobUrl)}`.

Utils/DecryptBlobFile/DecryptBlobFile.cs
```diff
@@ -0,0 +1,76 @@
...
+                        return outRecord;
+                    }
+
+                    CloudBlockBlob blob = new CloudBlockBlob(new Uri(blobUrl + sasToken));
```
bleroy, line 57: No validation on `sasToken`?

Utils/DecryptBlobFile/README.md
```diff
@@ -0,0 +1,125 @@
...
+            },
+            "errors": null,
+            "warnings": null
+        }
+    ]
+}
+```
+
+## Sample Skillset Integration
+
```
bleroy, line 69: Is it a relevant scenario at all, and is it worth pointing out what to do if not all files are encrypted, and is there a way in that case to avoid the round-trip to the custom skill with a conditional or something similar?

bleroy (changes requested):

bleroy:
This is super-useful, thanks!

Utils/DecryptBlobFile/DecryptBlobFile.cs
Careyjmac: I had copied this from the SplitImage power skill and it doesn't do any validation (though it does do some checking to make sure the url doesn't already contain the sas, which i can add here if we think it is useful). Do you know how to properly validate a sas token?  IMO we don't need to validate it isn't null or empty because if the blob is public then we don't need it.  Though i can't imagine why your blob would be public if it is also encrypted :P

Utils/DecryptBlobFile/README.md
Careyjmac: So I don't think this is relevant; even if it is possible that a container contains some encrypted blobs and some not, because since we are requesting the user set "dataToExtract": "storageMetadata", they will need this skill and the document extraction skill to get the content for the unencrypted blobs as well.  I should validate that this skill is able to download blobs that are not encrypted though, I don't see why it wouldn't be able to but ill run a manual test to make sure.

Utils/DecryptBlobFile/DecryptBlobFile.cs
Careyjmac: I made the combining code from the SplitImage skill a helper and refactored SplitImage and AnalyzeForm as well as this new skill to use it, hopefully that is sufficient.

Tests/PowerSkillTests/GeoPointFromNameTests/GeoPointFromNameTests.cs
```diff
@@ -20,17 +21,22 @@ public static void Setup(TestContext context)
             WebApiSkillHelpers.TestMode = true;
             WebApiSkillHelpers.TestWww = req =>
             {
-                string query = req.RequestUri.ParseQueryString()["query"];
+                string query = req.RequestUri.ParseQueryString()["q"];
```
Careyjmac, line 24: FYI, there was a test broken because of the recent API contract change not being applied here, so I just went ahead and fixed it, all tests are passing now :)

Utils/DecryptBlobFile/README.md
Careyjmac: I confirmed that it all works just fine if some files are encrypted and others are not, it is still able to index both of them.

Tests/PowerSkillTests/GeoPointFromNameTests/GeoPointFromNameTests.cs
bleroy: Thanks!

bleroy (approved):

PR #56, opened 2021-02-16
Improved readability of Readme page

vkurpad:
## Purpose

* ...

```
[ ] Yes
[ ] No
```

## Pull Request Type
Cleaned up readme to better describe the skill types and how they are implemented

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


bleroy:
I really like how the new table of skills is looking. How does that look on a phone however?

bleroy:
Some of the badges appear really small and barely readable:
image

bleroy:
Can you move the "up for grabs" section to the end, just before references?

bleroy:
The order of the skills seems like it arbitrarily grew with time (because that's exactly what happened). We might want to take this opportunity to sort them in a more rational order. Hello world should always be first, but maybe after that we sort by skill type and then alphabetical order?

PR #67, opened 2021-06-18
Healthcare TA Powerskill

jonathanserbent:
## Purpose
This PR adds a powerskill which utilizes the Healthcare Text Analytics API.

```
[ ] Yes
[x] No
```

## Pull Request Type
```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone https://github.com/jonathanserbent/azure-search-power-skills.git
cd azure-search-power-skills
git checkout HealthcareTA
npm install
```
TODO: unsure about this last `npm install` command. It was included in the PR template but I don't think it applies since we are using nuget packages, not npm.

* Test the code

TODO: Finish writing tests and add steps here.
```
```

## What to Check
* `HealthcareTA.cs` implementation
* Tests in `HealthcareTATests.cs` are extensive
  * TODO: Finish implementing them
* Check that ARM file deploys nicely
* Provide feedback on readme

## Other Information


HealthcareTA/.gitignore
```diff
@@ -0,0 +1,264 @@
+## Ignore Visual Studio temporary files, build results, and
```
bleroy, line 1: This is very noisy. Can you boil it down to what you really need and that's different from the one in the repo's root? And if your changes make sense on a global level, just add them to the root.

HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+        private static async Task ExtractEntityData(AsyncPageable<AnalyzeHealthcareEntitiesResultCollection> pages, WebApiResponseRecord outRecord)
+        {
+            await foreach (AnalyzeHealthcareEntitiesResultCollection documentsInPage in pages)
+            {
+                foreach (AnalyzeHealthcareEntitiesResult entitiesInDoc in documentsInPage)
+                {
+                    if (!entitiesInDoc.HasError)
+                    {
+                        outRecord.Data["entities"] = entitiesInDoc.Entities;
+                        outRecord.Data["relations"] = entitiesInDoc.EntityRelations;
```
bleroy, line 124: Aren't you getting data from only the last entities on the last page?

bleroy (changes requested):
Thanks for the contribution. You'll also want to add a readme page, a link to that readme from the top-level readme, and deployment files. Take a look at other C# skills for reference.

HealthcareTA/.gitignore
jonathanserbent: This gitignore file was auto-generated by VS when creating the project, but isn't needed, so I removed it

Text/HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using System.Collections.Generic;
+using AzureCognitiveSearch.PowerSkills.Common;
+using Azure.AI.TextAnalytics;
+using Azure;
+
+namespace AzureCognitiveSearch.PowerSkills.Text.HealthcareTA
+{
```
Careyjmac, line 16: Can we call this "TextAnalyticsForHealth" since that is the official name they use? Make sure you update the project and folder name too.

Text/HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+            string skillName = executionContext.FunctionName;
+            IEnumerable<WebApiRequestRecord> requestRecords = WebApiSkillHelpers.GetRequestRecords(req);
+            if (requestRecords == null)
+            {
+                return new BadRequestObjectResult($"{skillName} - Invalid request record array.");
+            }
+
+            // Get Endpoint and access key from App Settings
+            string apiKey = Environment.GetEnvironmentVariable(healthcareApiEnvKey, EnvironmentVariableTarget.Process);
+            string apiEndpoint = Environment.GetEnvironmentVariable(healthcareApiEnvEndpoint, EnvironmentVariableTarget.Process);
```
Careyjmac, line 39: >string apiEndpoint = Environment.GetEnvironmentVariable(healthcareApiEnvEndpoint, EnvironmentVariableTarget.Process); [](http://example.com/codeflow?start=12&length=117)

Do we need the endpoint to be in an environment variable? It should be pretty standard and able to be hardcoded right? You might have copied this from the form recognizer example because that one does have a custom endpoint (since a customer has to pretain a form recognizer model) but this one shouldn't.

Text/HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+            if (apiKey == null || apiEndpoint == null)
+            {
+                return new BadRequestObjectResult($"{skillName} - Healthcare Text Analytics API key or endpoint is missing. Make sure to set them in the Environment Variables.");
+            }
+
+            var client = new TextAnalyticsClient(new Uri(apiEndpoint), new AzureKeyCredential(apiKey));
+
+            // Get a custom timeout from the header, if it exists. If not use the default timeout.
+            int timeout;
+            if (!int.TryParse(req.Headers["SkillTimeout"].ToString(), out timeout))
```
Careyjmac, line 49: Make sure you make a note when you document this that it needs to be in seconds.

Careyjmac: Also, we should validate to make sure its between 0 and 230 (your default) since anything else wouldn't work

Text/HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                timeout = defaultTimeout;
+            }
+            var timeoutMiliseconds = timeout * 1000;
+
+            WebApiSkillResponse response = await WebApiSkillHelpers.ProcessRequestRecordsAsync(skillName, requestRecords,
+                async (inRecord, outRecord) => {
+                    // Prepare analysis operation input
+                    var document = inRecord.Data["document"] as string;
+                    var options = new AnalyzeHealthcareEntitiesOptions { };
```
Careyjmac, line 59: >var options = new AnalyzeHealthcareEntitiesOptions { }; [](http://example.com/codeflow?start=20&length=55)

What kind of options are available here? Anything we might want to expose via a header config?

Text/HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+                    if (document.Length >= maxCharLength)
+                    {
+                        outRecord.Warnings.Add(new WebApiErrorWarningContract
+                        {
+                            Message = $"Healthcare Text Analytics Error: The submitted document was over {maxCharLength} characters. It has been truncated to fit this requirement."
+                        });
+                        document = document.Substring(0, maxCharLength);
+                    }
```
Careyjmac, line 67: According to the TA docs, this should be:
"5,120 characters as measured by StringInfo.LengthInTextElements."

I can show you how to use the StringInfo API to also do the substring if you want, though its pretty easy. Basically what this does is better handles documents that have like emojis in them

Text/HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+                    List<string> batchInput = new List<string>()
+                    {
+                        document
+                    };
```
Careyjmac, line 71: I do think we should at the very least make a follow up work item to see if we can batch multiple documents into a single request here

Text/HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+
+                    // start analysis process
+                    var timer = System.Diagnostics.Stopwatch.StartNew();
+                    AnalyzeHealthcareEntitiesOperation healthOperation = await client.StartAnalyzeHealthcareEntitiesAsync(batchInput, "en", options);
```
Careyjmac, line 75: >"en" [](http://example.com/codeflow?start=134&length=4)

We should make the language configurable in the input (with a default of English)

Text/HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+                    var healthOperationTask = healthOperation.WaitForCompletionAsync().AsTask();
+
+                    if (await Task.WhenAny(healthOperationTask, Task.Delay(timeoutMiliseconds)) == healthOperationTask)
```
Careyjmac, line 78: >if (await Task.WhenAny(healthOperationTask, Task.Delay(timeoutMiliseconds)) == healthOperationTask) [](http://example.com/codeflow?start=20&length=99)

So keep in mind a feww things here:

1. That max timeout of 230 seconds is the max overall function runtime. So if you wait until now to start that timer, it'll probably be too late and the function will just crash without sending the nice error message we want. We should probably add a 5 seconds buffer or something to counteract that.
2. You are measuring the timeout per request, but you might be sending multiple requests per input depending on how the customer configured the skill, and as I mentioned the timeout needs to be for the function in its entirety.

Text/HealthcareTA/HealthcareTA.cs
jonathanserbent: I think that the TextAnalytics API *does* have a constant endpoint, but when allocating a resource for a text analytics account, the user also gets an endpoint within a unique subdomain. I figured since most people will be deploying this with the arm template we can use the endpoint with the subdomain, that way if the endpoint changes, it will automatically get updated.
I could also include a check where if there is not endpoint it defaults to the constant endpoint from the TextAnalytics API.

Text/HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+                        // Task Completed, now lets process the result.
+                        outRecord.Data["status"] = healthOperation.Status.ToString();
+                        if (healthOperation.Status != TextAnalyticsOperationStatus.Succeeded)
+                        {
+                            // The operation was not a success
+                            outRecord.Warnings.Add(new WebApiErrorWarningContract
+                            {
+                                Message = "Healthcare Text Analytics Error: Health Operation returned a non-succeeded status."
+                            });
+                        }
```
Careyjmac, line 89: Two things:

1. This should probably be an error, not a warning
2. We should probably have a separate error message for the timeout case, since the operation might still eventually complete successfully and therefore the customer might get charged for it. Perhaps we can include the operation id in the error in that case so the customer can follow up separately if they want?

Text/HealthcareTA/HealthcareTA.cs
Careyjmac, line 75: >await client.StartAnalyzeHealthcareEntitiesAsync(batchInput, "en", options); [](http://example.com/codeflow?start=73&length=76)

Oh something I just realized, this API supports passing a cancellationToken, so we might be able to just use that to control the timeout. Let's chat about what that means later today.

[https://docs.microsoft.com/en-us/dotnet/api/azure.ai.textanalytics.textanalyticsclient.startanalyzehealthcareentitiesasync?view=azure-dotnet-preview#Azure_AI_TextAnalytics_TextAnalyticsClient_StartAnalyzeHealthcareEntitiesAsync_System_Collections_Generic_IEnumerable_System_String__System_String_Azure_AI_TextAnalytics_AnalyzeHealthcareEntitiesOptions_System_Threading_CancellationToken](https://docs.microsoft.com/en-us/dotnet/api/azure.ai.textanalytics.textanalyticsclient.startanalyzehealthcareentitiesasync?view=azure-dotnet-preview#Azure_AI_TextAnalytics_TextAnalyticsClient_StartAnalyzeHealthcareEntitiesAsync_System_Collections_Generic_IEnumerable_System_String__System_String_Azure_AI_TextAnalytics_AnalyzeHealthcareEntitiesOptions_System_Threading_CancellationToken)_

HealthcareTA/HealthcareTA.cs
Careyjmac: Good catch, he's overwriting instead of appending each time. @jonathanserbent make sure to fix

HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+                });
+
+            return new OkObjectResult(response);
+        }
+
+        private static async Task ExtractEntityData(AsyncPageable<AnalyzeHealthcareEntitiesResultCollection> pages, WebApiResponseRecord outRecord)
+        {
+            await foreach (AnalyzeHealthcareEntitiesResultCollection documentsInPage in pages)
+            {
+                foreach (AnalyzeHealthcareEntitiesResult entitiesInDoc in documentsInPage)
```
Careyjmac: >AnalyzeHealthcareEntitiesResult [](http://example.com/codeflow?start=25&length=31)

This does have a Warnings field so we should add those to the output as well.

Text/HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+                });
+
+            return new OkObjectResult(response);
+        }
+
+        private static async Task ExtractEntityData(AsyncPageable<AnalyzeHealthcareEntitiesResultCollection> pages, WebApiResponseRecord outRecord)
+        {
+            await foreach (AnalyzeHealthcareEntitiesResultCollection documentsInPage in pages)
+            {
+                foreach (AnalyzeHealthcareEntitiesResult entitiesInDoc in documentsInPage)
```
Careyjmac, line 119: >AnalyzeHealthcareEntitiesResult [](http://example.com/codeflow?start=25&length=31)

This does have a Warnings field so we should add those to the output as well.

Text/HealthcareTA/HealthcareTA.cs
Careyjmac: Nevermind the second note, I see you already have that below and I just misunderstand. So yea just make them errors instead of warnings imo.

Text/HealthcareTA/HealthcareTA.cs
```diff
@@ -0,0 +1,136 @@
...
+                        // Timeout
+                        outRecord.Warnings.Add(new WebApiErrorWarningContract
+                        {
+                            Message = "Healthcare Text Analytics Error: The Text Analysis Operation took too long to complete."
+                        });
+                    }
+
+                    // Record how long this task took to complete.
+                    timer.Stop();
+                    outRecord.Data["timeToComplete"] = timer.Elapsed.TotalSeconds.ToString();
```
Careyjmac, line 107: I don't think we'll actually want to check this in since customers likely won't care. We can add a log for it in the Azure Function for those that do, it just doesn't need to be an output if that makes sense.

Text/HealthcareTA/azuredeploy.json
```diff
@@ -0,0 +1,143 @@
...
+      ],
+      "type": "String",
+      "metadata": {
+        "description": "Storage Account type"
+      }
+    }
+  },
+  "variables": {
+    "repoURL": "https://github.com/jonathanserbent/azure-search-power-skills/",
+    "repoBranch": "HealthcareTA",
```
Careyjmac, line 28: Make sure to update these to match the upstream before we merge

Text/HealthcareTA/azuredeploy.json
```diff
@@ -0,0 +1,143 @@
...
+              "value": "[variables('functionProject')]"
+            },
+            {
+              "name": "HEALTHCARE_API_KEY",
+              "value": "TODO"
+            },
+            {
+              "name": "HEALTHCARE_API_ENDPOINT",
+              "value": "TODO"
+            }
```
Careyjmac, line 115: At least for the key, you should be able to do this using the listkeys function. There is an example of this in the JFK sample:

[https://github.com/microsoft/AzureSearch_JFK_Files/blob/9ba42347a6285ce0b73b5bcb935e8c84852f9cfc/azuredeploy.json#L250](https://github.com/microsoft/AzureSearch_JFK_Files/blob/9ba42347a6285ce0b73b5bcb935e8c84852f9cfc/azuredeploy.json#L250)

Another useful doc for this:
[https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions)

Text/HealthcareTA/HealthcareTA.cs
Careyjmac: Gotcha, yea let's just add a default to the constant one then

bleroy: This might also help with testing. Variable + default sounds good.

Text/HealthcareTA/azuredeploy.json
Careyjmac: https://docs.microsoft.com/en-us/rest/api/cognitiveservices/accountmanagement/accounts

jonathanserbent: Got that implemented and pushed. The docs were super helpful, thank you!

Text/HealthcareTA/HealthcareTA.cs
Careyjmac: https://docs.microsoft.com/en-us/dotnet/api/azure.ai.textanalytics.analyzehealthcareentitiesoptions?view=azure-dotnet-preview

jonathanserbent: Synched with Carey on this, no options we want to expose in the header.

Text/HealthcareTA/azuredeploy.json
```diff
@@ -0,0 +1,143 @@
...
+              "name": "AzureWebJobsDashboard",
+              "value": "[concat('DefaultEndpointsProtocol=https;AccountName=', variables('storageAccount'), ';AccountKey=', listKeys(resourceId('Microsoft.Storage/storageAccounts', variables('storageAccount')), '2021-04-01').keys[0].value)]"
+            },
+            {
+              "name": "AzureWebJobsStorage",
+              "value": "[concat('DefaultEndpointsProtocol=https;AccountName=', variables('storageAccount'), ';AccountKey=', listKeys(resourceId('Microsoft.Storage/storageAccounts', variables('storageAccount')), '2021-04-01').keys[0].value)]"
+            },
+            {
+              "name": "FUNCTIONS_EXTENSION_VERSION",
+              "value": "~2"
```
Careyjmac, line 94: Change this to "~3" to make the arm template deployment work. This is because your csproj references v3 on line 4, so right now you are creating a v2 function but trying to deploy v3 code to it. 

The other power skills have the correct v2 matching hence why this was a copy and paste error. Sorry for the confusion there!

Text/HealthcareTA/HealthcareTA.cs
jonathanserbent: Added a default for the endpoint

HealthcareTA/HealthcareTA.cs
jonathanserbent: Just started looking into this, I need to read up more on the docs for the TA library, but currently the results I have checked all have just one document (which makes sense because for each request we are only sending one), but it also only ever has one collection of entities/entity relations.

Careyjmac: Went over this with Jonathan, the TA API doesn't make that much sense but we assume they made it pageable in the SDK just in case they make the actual API pageable in the future. With Jonathan's current non-batching implementation this should work as is, but I did ask him to just make sure we create a list that we append to just in case.

jonathanserbent: Implemented and pushed the change Carey Mentioned above.

Text/HealthcareTA/HealthcareTA.cs
jonathanserbent: Updated to check for timeout across all requests

Text/HealthcareTA/HealthcareTA.cs
jonathanserbent: Switched those to errors

Text/HealthcareTA/HealthcareTA.cs
jonathanserbent: done and done 👍

Text/HealthcareTA/HealthcareTA.cs
jonathanserbent: done and done 👍

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
```diff
@@ -0,0 +1,189 @@
...
+                            outRecord.Warnings.Add(new WebApiErrorWarningContract
+                            {
+                                Message = $"Healthcare Text Analytics Warning: {w.WarningCode}. Error Message: {w.Message}"
+                            });
+                        }
+                    }
+                }
+            }
+            outRecord.Data[$"entities"] = entities;
+            outRecord.Data[$"relations"] = entities;
```
Careyjmac, line 186: >entities; [](http://example.com/codeflow?start=43&length=9)

Typo: relations?

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
```diff
@@ -0,0 +1,189 @@
...
+using Azure.AI.TextAnalytics;
+using Azure;
+using System.Globalization;
+
+namespace AzureCognitiveSearch.PowerSkills.Text.TextAnalyticsForHealth
+{
+    public static class TextAnalyticsForHealth
+    {
+        public static readonly string textAnalyticsApiEndpointSetting = "TEXT_ANALYTICS_API_ENDPOINT";
+        public static readonly string defaultTextAnalyticsEndpoint = "https://centralus.api.cognitive.microsoft.com";
```
Careyjmac, line 21: >centralus [](http://example.com/codeflow?start=78&length=9)

Oh, so the default endpoint is regional? Yea perhaps we shouldn't have a default endpoint then, since this will depend on what region the customer creates their resources in with the ARM template

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
```diff
@@ -0,0 +1,189 @@
...
+            if (!int.TryParse(req.Headers["SkillTimeout"].ToString(), out timeout))
+            {
+                timeout = defaultTimeout;
+            }
+            timeout = Math.Clamp(timeout - timeoutBuffer, 0, maxTimeout);
+            var timeoutMiliseconds = timeout * 1000;
+            var timeoutTask = Task.Delay(timeoutMiliseconds);
+
+            // Get a custom default language, if none is provided, use english
+            string defaultLanguage = req.Headers.ContainsKey("DefaultLanguage")? req.Headers["DefaultLanguage"].ToString(): "en";
```
Careyjmac, line 65: >DefaultLanguage [](http://example.com/codeflow?start=62&length=15)

Let's call this "defaultLanguageCode" (with the camel instead of pascal casing) to match what we call it for built in skills

Text/TextAnalyticsForHealth/azuredeploy.json
```diff
@@ -0,0 +1,143 @@
...
+        }
+      ]
+    }
+  ],
+  "outputs": {
+    "Azure Function Site Name": {
+      "type": "String",
+      "value": "[variables('functionAppName')]"
+    }
+  }
```
Careyjmac, line 142: I would love to see if there is a way to output the function auth code in ARM templates now (it wasn't possible before which is why other power skills don't do it). Not a big deal if you want to just file a follow up work item for it though

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <ProjectReference Include="..\..\Common\Common.csproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Update="host.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="local.settings.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
+    </None>
```
Careyjmac, line 20: We should probably check in the local.settings.json file if we want to include it here (i.e. if it is useful for local debugging, include it with the needed environment variable keys but with empty values), or otherwise remove it from the csproj

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
```diff
@@ -0,0 +1,189 @@
+using System;
```
Careyjmac, line 1: Add copyright note to the beginning of this file like we have in other power skills:

// Copyright (c) Microsoft. All rights reserved.  
// Licensed under the MIT License. See LICENSE file in the project root for full license information.

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
```diff
@@ -0,0 +1,189 @@
...
+            }
+            if (apiKey == null)
+            {
+                return new BadRequestObjectResult($"{skillName} - Healthcare Text Analytics API key is missing. Make sure to set it in the Environment Variables.");
+            }
+            var client = new TextAnalyticsClient(new Uri(apiEndpoint), new AzureKeyCredential(apiKey));
+
+            // Get a custom timeout from the header, if it exists. If not use the default timeout.
+            int timeout;
+            if (!int.TryParse(req.Headers["SkillTimeout"].ToString(), out timeout))
```
Careyjmac, line 56: >"SkillTimeout" [](http://example.com/codeflow?start=42&length=14)

Call this just "timeout" to match the WebApiSkill definition naming

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
```diff
@@ -0,0 +1,189 @@
...
+            string defaultLanguage = req.Headers.ContainsKey("DefaultLanguage")? req.Headers["DefaultLanguage"].ToString(): "en";
+
+            WebApiSkillResponse response = await WebApiSkillHelpers.ProcessRequestRecordsAsync(skillName, requestRecords,
+                async (inRecord, outRecord) => {
+                    if (timeoutTask.IsCompleted)
+                    {
+                        // The time limit for all the skills has been met
+                        outRecord.Errors.Add(new WebApiErrorWarningContract
+                        {
+                            Message = "Healthcare Text Analytics Error: The Text Analysis Operation took too long to complete."
```
Careyjmac, line 74: >Healthcare Text Analytics [](http://example.com/codeflow?start=39&length=25)

Let's update all of the errors/warnings to say TextAnalyticsForHealth as well

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
```diff
@@ -0,0 +1,189 @@
...
+            }
+            var client = new TextAnalyticsClient(new Uri(apiEndpoint), new AzureKeyCredential(apiKey));
+
+            // Get a custom timeout from the header, if it exists. If not use the default timeout.
+            int timeout;
+            if (!int.TryParse(req.Headers["SkillTimeout"].ToString(), out timeout))
+            {
+                timeout = defaultTimeout;
+            }
+            timeout = Math.Clamp(timeout - timeoutBuffer, 0, maxTimeout);
```
Careyjmac, line 60: >timeout = Math.Clamp(timeout - timeoutBuffer, 0, maxTimeout); [](http://example.com/codeflow?start=12&length=61)

I'm still a little wary about doing the buffer subtraction inside the function. Say the customer requests 300 seconds, then this will come out to 

Math.Clamp(300-5,0,230)=Math.Clamp(295,0,230)=230 

instead of 225 like to actually it actually should be to give the buffer. Also a timeout of 0 isn't very useful tbh. Let's chat at our sync quickly on this.

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
```diff
@@ -0,0 +1,189 @@
...
+                    if (!inRecord.Data.ContainsKey("document"))
+                    {
+                        outRecord.Errors.Add(new WebApiErrorWarningContract
+                        {
+                            Message = "Healthcare Text Analytics Error: The skill request did not contain 'document' in the input."
+                        });
+                        return outRecord;
+                    }
+                    var document = inRecord.Data["document"] as string;
+                    var language = inRecord.Data.ContainsKey("language") ? inRecord.Data["language"] as string : defaultLanguage;
```
Careyjmac, line 89: More naming nits: let's do "document" -> "text" and "language" -> "languageCode", again just to better line up with our built in TA skills.

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
```diff
@@ -0,0 +1,189 @@
...
+                        document = docInfo.SubstringByTextElements(0, maxCharLength);
+                    }
+
+                    var options = new AnalyzeHealthcareEntitiesOptions { };
+                    List<string> batchInput = new List<string>()
+                    {
+                        document
+                    };
+
+                    // start analysis process TODO error check
```
Careyjmac, line 107: >TODO error check [](http://example.com/codeflow?start=46&length=16)

Is this done and we should remove the TODO?

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
```diff
@@ -0,0 +1,189 @@
...
+            // all output into these two collections.
+            var entities = new List<Object>();
+            var relations = new List<Object>();
+            await foreach (AnalyzeHealthcareEntitiesResultCollection page in pages)
+            {
+                foreach (AnalyzeHealthcareEntitiesResult document in page)
+                {
+                    if (!document.HasError)
+                    {
+
```
Careyjmac, line 162: super ultra nit: remove extra whitespace

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.csproj
jonathanserbent: I am very unsure of what the `.csproj` file represents or how it was changed. I think this was changed by VS. I'm happy to change it to whatever you think it best.

Careyjmac: I would say just go ahead and commit the local.settings.json file, and then you should be good. Like I said it can be useful for local debugging anyways.

Text/TextAnalyticsForHealth/azuredeploy.json
jonathanserbent: I was able to get it to output the full URI for the skill. should be resolved but for some reason this block doesn't show as `outdated`. Let me know if you can't find that change and I will make sure it gets there. :)

Careyjmac: Wooo so awesome! This will be sooooo easy for customers to use now :D We need to update the other power skills with all of these improvements @bleroy

README.md
```diff
@@ -36,6 +36,7 @@ This project provides the following custom skills:
...
 | [HelloWorld](Template/HelloWorld/README.md) | A minimal skill that can be used as a starting point or template for your own skills. | Template | ![C#](https://img.shields.io/badge/language-C%23-brightgreen) | ![functions](https://img.shields.io/badge/deploy-Functions-blue) | ARM Template |
 | [PythonFastAPI](Template/PythonFastAPI/README.md) | A production web server and api scaffold for a python power skill | Template | ![python](https://img.shields.io/badge/language-python-orange) | ![docker](https://img.shields.io/badge/deploy-Docker-blueviolet) | Terraform template |
+| [TextAnalyticsForHealth](Text/TextAnalyticsForHealth/README.md) | A wrapper for the Text Analytics for Health API | Template | ![C#](https://img.shields.io/badge/language-C%23-brightgreen) | ![functions](https://img.shields.io/badge/deploy-Functions-blue) | ARM Template |
```
Careyjmac, line 39: >Template [](http://example.com/codeflow?start=118&length=8)

I think this should be "Text" since its not a template for something but rather does a specific text analysis. Also when you change that move it up to be grouped with the other power skills of type "Text"

Text/TextAnalyticsForHealth/README.md
```diff
@@ -0,0 +1,118 @@
+﻿---
+page_type: sample
+languages:
+- csharp
+products:
+- azure
+- azure-search
+- azure-cognitive-services
+name: Bing Entity Search sample skill for cognitive search
+description: This custom skill finds rich and structured information about public figures, locations, or organizations.
```
Careyjmac, line 10: Update these :)

Text/TextAnalyticsForHealth/README.md
```diff
@@ -0,0 +1,118 @@
...
+azureDeploy: https://raw.githubusercontent.com/Azure-Samples/azure-search-power-skills/main/Text/TextAnalyticsForHealth/azuredeploy.json
+---
+
+# TextAnalyticsForHealth
+
+This custom skill finds rich and structured information about public figures, locations, or organizations.
```
Careyjmac, line 16: And this description

Text/TextAnalyticsForHealth/README.md
```diff
@@ -0,0 +1,118 @@
...
+    ]
+}
+```
+The skill also accepts options which can be sent as headers in the request.
+
+| Name | Description | Default |
+| ---- | ----------- | ------- |
+| `timeout` | Allows you to set a timeout (in seconds) which will cause the skill to return early, so it doesn't timeout in the Cognitive Search Pipeline. | `230` |
+| `defaultLanguage` | Allows you to set a default language across all documents in case a language is not provided in the input | `"en"` |
+
```
Careyjmac, line 62: Should probably move this to be under the "Sample skillset integration" options below, and show how you can set them via the httpHeaders parameter of the skill

Text/TextAnalyticsForHealth/README.md
```diff
@@ -0,0 +1,118 @@
...
+{
+    "values": [
+        {
+            "recordId": "foobar2",
+            "data":
+            {
+                "text":  "100mg Ibuprofin"
+            }
+        },
+        {
```
Careyjmac, line 45: super nit: fix indentation

Text/TextAnalyticsForHealth/README.md
```diff
@@ -0,0 +1,118 @@
...
+                "relations": [
+                    {...}
+                ]
+            },
+            "warnings": [],
+            "errors": []
+        },
+        {...}
+    ]
+}
```
Careyjmac, line 84: It would be great if we could expand this to show the full output based on the sample inputs you provided, I know it can be tedious but it can help customers if they want to select specific things via field mappings later on (i.e. if they need to do "/document/entities/*/name" or something)

Text/TextAnalyticsForHealth/README.md
Careyjmac: And this is where you can probably include a link to the TA for health docs

Text/TextAnalyticsForHealth/README.md
```diff
@@ -0,0 +1,118 @@
...
+In order to use this skill in a cognitive search pipeline, you'll need to add a skill definition to your skillset. **If you deploy using the ARM template you can copy the URI from the output of the template.**
+Here's a sample skill definition for this example (inputs and outputs should be updated to reflect your particular scenario and skillset environment):
+
+```json
+{
+    "@odata.type": "#Microsoft.Skills.Custom.WebApiSkill",
+    "description": "Our new Bing entity search custom skill",
+    "context": "/document",
+    "uri": "[AzureFunctionEndpointUrl]/api/entity-search?code=[AzureFunctionDefaultHostKey]",
+    "batchSize": 1,
```
Careyjmac, line 98: Update description and uri (make sure the api path is the one for your function) and then also add the timeout and httpHeaders parameters (just use the default values that you use in your code for the purpose of the sample skill definition)

Careyjmac: Oh I did just realize httpHeaders is on line 116, I would remove it from there and move it up here and include the default settings for your special headers as well

Text/TextAnalyticsForHealth/README.md
```diff
@@ -0,0 +1,118 @@
...
+    "inputs": [
+        {
+            "name": "document",
```
Careyjmac, line 101: >"document", [](http://example.com/codeflow?start=20&length=11)

Change to "text"

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
Careyjmac: Basically what I'm saying is if this is the case, we should just get rid of the default endpoint altogether. Sorry for the confusion and extra work earlier. :P

Tests/PowerSkillTests/TextAnalyticsForHealthTests/TextAnalyticsForHealthTests.cs
```diff
@@ -0,0 +1,80 @@
...
+using System;
+using Microsoft.AspNetCore.Http.Internal;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+
+namespace AzureCognitiveSearch.PowerSkills.Tests.TextAnalyticsForHealthTests
+{
+    /** TODO: Need to come back and find a way to test the Function better. Currently the tests fail when using the
+     * helper functions, which we believe is because of how the Text Analytics SDK sends requests to their endpoint.
+     */
```
Careyjmac, line 19: Make sure to create a follow up work item for this as well :)

Text/HealthcareTA/HealthcareTA.cs
jonathanserbent: Created a work item for this: https://msdata.visualstudio.com/Azure%20Search/_workitems/edit/1251209

Text/HealthcareTA/HealthcareTA.cs
jonathanserbent: Created a work item for this: https://msdata.visualstudio.com/Azure%20Search/_workitems/edit/1251217

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
jonathanserbent: So right now the endpoint will be set in the settings by the ARM template, so if they are deploying it from there it should all be gravy. If they deploy it on their own, or for some reason the ARM template fails, and it causes this setting to not be set, would we rather have the function error out on all requests or use a regional endpoint which is less efficient?

Text/TextAnalyticsForHealth/README.md
jonathanserbent: I think I got this. lmk if it looks right when I commit/push

Text/TextAnalyticsForHealth/README.md
```diff
@@ -70,10 +64,63 @@ The skill also accepts options which can be sent as headers in the request.
...
+                        "subCategory": null,
+                        "confidenceScore": 1.0,
+                        "offset": 0,
+                        "length": 5,
+                        "dataSources": [],
+                        "assertion": null,
+                        "normalizedText": null
+                    },
+                    {
+                        "text": "Ibuprofin",
```
bleroy, line 79: It's Ibuporofen, not Ibuprofin.

jonathanserbent: I've said it before and I'll say it again, my greatest flaw is that I cannot spell things correctly. Thanks for the catch lol

Text/TextAnalyticsForHealth/README.md
Careyjmac: Yup fixed!

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.cs
Careyjmac: The problem isn't that the regional endpoint is less efficient, its that I'm pretty sure it straight up won't work if the region of the key you are using isn't created in Central US. Hence why its proabably just better to require the endpoint in the function settings as you originally did, especially since for 99% of users the ARM template will just take care of it for them anyways.

Text/TextAnalyticsForHealth/TextAnalyticsForHealth.csproj
Careyjmac: Oh huh, weird. @bleroy we should probably look into why other csprojs still reference it. But if they all still work then that is fine for you to leave as is for now Jonathan.

Text/TextAnalyticsForHealth/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+In order to use this skill in a cognitive search pipeline, you'll need to add a skill definition to your skillset. **If you deploy using the ARM template you can copy the URI from the output of the template.**  
+The skill also accepts options which can be sent as headers in the request.
+
+| Header Name | Description | Default |
+| ----------- | ----------- | ------- |
+| `timeout` | Allows you to set a timeout (in seconds) which will cause the skill to return early, so it doesn't timeout in the Cognitive Search Pipeline. | `230` |
+| `defaultLanguageCode` | Allows you to set a default language across all documents in case a language is not provided in the input | `en` |
+
+**It is important to note** the difference between the timeout option in the header, and the timeout option/behavior when defining a skillset. If a request to a web skill takes longer than the timeout defined in the skillset, it will simply drop the request and move on. With the timeout header option, it allows users to define how long the function should wait before it returns the results it has so far, and returns errors for the documents it has not yet gotten to or not yet received results on.
```
Careyjmac, line 144: I love this clarification, thank you for adding it!

Careyjmac (approved):

bleroy (approved):
LGTM

PR #99, opened 2022-07-31
Refactor custom text powerskills and improve testing

ramezgerges:
## Purpose

* Custom NER and custom text classification powerskills were refactored to be more unit-testable and overall cleaner.
* Unit tests were added and E2E tests were rewritten to be more robust.
* More details about how Azure search indexes can be useful were added in the README's.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Refactoring (no functional changes, no api changes)
[x] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone https://github.com/Azure-Samples/azure-search-power-skills.git
cd azure-search-power-skills/Text/CustomNER # or CustomTextClassifier
python -m pip install -r requirements.txt 
```

* Test the code

```
 python -m unittest
```

## What to Check
* ...


Text/CustomTextClassifier/README.md
```diff
@@ -225,11 +227,142 @@ Finally, the indexer ties everything together. The indexer needs to be setup up
 }
 ```
 
+#### Query the index
+
+Now that we have a packed index, we can use the capabilities of the index to make useful queries. For example, using the sample data for scientific papers from the custom text classification documentation (found [here](https://github.com/Azure-Samples/cognitive-services-sample-data-files/tree/master/language-service/Custom%20text%20classification)), we can query the index to find all load agreements on a specific date. For example,
```
bleroy, line 232: "populated" rather than "packed"?

bleroy (approved):

Text/CustomTextClassifier/README.md
ramezgerges: I kept looking for "populated" for about 5 minutes. It only came to me after I pushed :).

ramezgerges:
Based on some input from team members inside Microsoft, I've simplified the README's and added ready-to-deploy zip files instead of forcing the user to create their own zips. Considering this doesn't cause breaking changes, and overall is similar in spirit to the initial commits in this PR (in that it's just general improvements and cleaning up), I've chosen to add these changes to this PR instead of opening a new one.

bleroy:
Thanks for the contribution!

PR #220, opened 2024-11-22
Fix main branch codebase for openai chat completion language model skill

amitkalay:
While testing out this code locally, I saw that it was broken in its current state. These code changes fix the codebase here so that things work right out of the box. I am following @farzad528 's example here - https://github.com/farzad528/azure-search-power-skills/tree/main/AzureOpenAICustomLanguageModelSkill


farzad528 (approved):
lgtm

PR #224, opened 2024-11-26
Consolidate chat completion custom skill code for Open-AI and non OpenAI language models

amitkalay:
Previously we drew a line for doing chat-completion inference tasks based on whether the language model was from OpenAI or different (llama, Phi, Cohere, etc.). I've now converged that code so that a user can use the exact same custom skill code to call whichever language model they prefer. All they have to do is modify their environment variables.

I tested this by running this code across text-summarization, entity recognition and image verbalization AND across different language model deployments (gpt-4o & phi 3-5)

### DIFF
--- a/ChatCompletionCustomSkill/README.md
+++ b/ChatCompletionCustomSkill/README.md
@@ -4,7 +4,7 @@
 
 Quickstart Guide:
 
-- This folder illustrates how to leverage Azure AI Studio to summarize a large piece of text using [custom web api skills](https://learn.microsoft.com/en-us/azure/search/cognitive-search-custom-skill-web-api).
+- This folder illustrates how to leverage Azure AI Studio with a deployed language model of your choice to do a chat-completion skill using [custom web api skills](https://learn.microsoft.com/en-us/azure/search/cognitive-search-custom-skill-web-api).
 
 - As a prerequisite to running this custom skill, you must first deploy a model to Azure. See [this guide](https://learn.microsoft.com/en-us/azure/ai-studio/how-to/deploy-models-openai) for reference.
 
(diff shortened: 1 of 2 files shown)


ChatCompletionCustomSkill/function_app.py
```diff
@@ -203,44 +207,47 @@ async def custom_skill(req: func.HttpRequest) -> func.HttpResponse:
...
+                    "temperature": config.temperature,
+                    "top_p": config.top_p,
+                    "max_tokens": config.max_tokens
+                }
+                # Call model with timeout
+                async with asyncio.timeout(config.timeout):
+                    vanilla_response = requests.post(endpoint, headers=headers, json=request_payload)
+                    vanilla_response.raise_for_status()  # Will raise an HTTPError if the HTTP request returned an unsuccessful status code
+                    vanilla_response_json = vanilla_response.json()
+                    response_text = vanilla_response_json['choices'][0]['message']['content']
```
bleroy, line 231: Do we need some checks in there if the structure is not what's expected or are we fine with the generic except below?

bleroy (approved):

ChatCompletionCustomSkill/README.md
```diff
@@ -4,7 +4,7 @@
 
 Quickstart Guide:
 
-- This folder illustrates how to leverage Azure AI Studio to summarize a large piece of text using [custom web api skills](https://learn.microsoft.com/en-us/azure/search/cognitive-search-custom-skill-web-api).
+- This folder illustrates how to leverage Azure AI Studio with a deployed language model of your choice to do a chat-completion skill using [custom web api skills](https://learn.microsoft.com/en-us/azure/search/cognitive-search-custom-skill-web-api).
```
steeler18, line 7: change to AI Foundry?

steeler18 (approved):

ChatCompletionCustomSkill/function_app.py
amitkalay: Good question... we don't really know the level of exception handling, validation, etc. that we'll need until we get much further along in the Native Skill arc. If there any gaps, I will address them!

ChatCompletionCustomSkill/README.md
amitkalay: whoops, I forgot about that name change!

